PR #31, opened 2019-12-12
Remove `config.toml` from hard-coded docker config

### LINKED ISSUE #27: Remove `config.toml` from hard-coded docker config
Instead of hardcoding the whole `config.toml` for seed/persistent peers. We can pass them directly as the command line arguments when running `bandd start`. That way, we have one less (quite big) file to maintain!

taobun:
fixed #27 

- Use config.toml from `bandd init` and add peer when start node


sorawit (approved):
Looks good.

PR #68, opened 2019-12-18
Implement code storage keeper and handler

### LINKED ISSUE #7: Implement code storage keeper and handler
Code storage keeper and handler allows the blockchain to process code storage messages. D3N should handle `MsgStoreCode` by keeping the code in a long term storage, using code hash as lookup key. `MsgDeleteCode` should remove the code from the storage. As simple as that!

- [ ] Implement keeper(s)
- [ ] Implement handler(s)
- [ ] Add tests

taobun:
fixed #7

### DIFF
--- a/chain/x/zoracle/handler_test.go
+++ b/chain/x/zoracle/handler_test.go
@@ -11,7 +11,6 @@ import (
 	"github.com/bandprotocol/d3n/chain/x/zoracle/internal/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	staking "github.com/cosmos/cosmos-sdk/x/staking"
-	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/stretchr/testify/require"
 	"github.com/tendermint/tendermint/libs/common"
 )
@@ -23,8 +22,8 @@ func TestRequestSuccess(t *testing.T) {
 	if err != nil {
 		fmt.Println(err)
 	}
-	codeHash := crypto.Keccak256(code)
 	sender := sdk.AccAddress([]byte("sender"))
+	codeHash := types.NewStoredCode(code, sender).GetCodeHash()
 
 	msg := types.NewMsgRequest(code, 5, sender)
 	got := handleMsgRequest(ctx, keeper, msg)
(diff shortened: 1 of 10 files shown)


chain/x/zoracle/handler_test.go
```diff
@@ -98,6 +97,94 @@ func TestReportInvalidValidator(t *testing.T) {
...
+
+	require.Equal(t, owner, sc.Owner)
+	require.Equal(t, code, sc.Code)
+}
+
+func TestStoreCodeFailed(t *testing.T) {
+	ctx, keeper := keep.CreateTestInput(t, false)
+	code := []byte("Code")
+
+	owner := sdk.AccAddress([]byte("owner"))
```
sorawit, line 130: Should we use realistic address here? (20 bytes?)

sorawit: Is it common in Cosmos-SDK tests to use "owner"?

chain/x/zoracle/internal/types/msgs.go
```diff
@@ -138,7 +138,7 @@ func (msg MsgStoreCode) ValidateBasic() sdk.Error {
 	}
 
 	if msg.Code == nil || len(msg.Code) == 0 {
-		return sdk.ErrUnknownRequest("Code must not be empty bytes")
+		return sdk.ErrUnknownRequest("Code must not be empty")
```
sorawit, line 141: Good catch!

sorawit (changes requested):
Logic. LGTM. Some questions to address in the review.

chain/x/zoracle/internal/types/code.go
```diff
@@ -0,0 +1,25 @@
...
+// NewStoredCode is a constructor of StoredCode
+func NewStoredCode(code []byte, owner sdk.AccAddress) StoredCode {
+	return StoredCode{
+		Code:  code,
+		Owner: owner,
+	}
+}
+
+// GetCodeHash is a function to calculate hash of actual code
+func (c StoredCode) GetCodeHash() []byte {
```
sorawit, line 23: Just to confirm. Even though this is pass by value, it's efficient because `c.Code` is a slice and Go does not do copy on slice, correct?

sorawit (changes requested):
Logic. LGTM. Some questions to address in the review.

chain/x/zoracle/internal/types/code.go
taobun: yep!

chain/x/zoracle/handler_test.go
taobun: It's common thing e.g. [bank_keeper](https://github.com/cosmos/cosmos-sdk/blob/master/x/bank/internal/keeper/keeper_test.go)

sorawit (approved):
Looks good.

chain/x/zoracle/internal/types/msgs.go
taobun: :)

PR #89, opened 2019-12-23
implement bandsv store code

### LINKED ISSUE #85: Implement bandsv store code

prin-r:
Fixes #85

### DIFF
--- a/chain/cmd/bandsv/main.go
+++ b/chain/cmd/bandsv/main.go
@@ -1,11 +1,18 @@
 package main
 
 import (
+	"encoding/hex"
 	"net/http"
+	"strconv"
 
-	"github.com/gin-gonic/gin"
-
+	"github.com/bandprotocol/d3n/chain/cmtx"
 	"github.com/bandprotocol/d3n/chain/x/zoracle"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/gin-gonic/gin"
+	"github.com/spf13/viper"
+	"github.com/tendermint/tendermint/crypto"
+	"github.com/tendermint/tendermint/crypto/secp256k1"
+	rpc "github.com/tendermint/tendermint/rpc/client"
 )
 
 type OracleRequest struct {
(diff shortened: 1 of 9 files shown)


chain/cmd/bandsv/main.go
```diff
@@ -18,7 +30,130 @@ type OracleRequestResp struct {
...
+	} `json:"response"`
+}
+
+const priv = "a96e62ed3955e65be3aaa3f12d87b6b5cf26039ecfa948dc5107a495418e5430"
+const port = "5001"
+const nodeURI = "http://localhost:26657"
+const queryURI = "http://localhost:1317"
+
+func GetTx(txHash string) (TxResult, error) {
+	_resp, err := http.Get(fmt.Sprintf("%s/tx?hash=0x%s", nodeURI, txHash))
```
sorawit: Let's not prefix the name with underscore?

chain/cmd/bandsv/main.go
```diff
@@ -18,7 +30,130 @@ type OracleRequestResp struct {
...
+		Height    string      `json:"height"`
+		Codespace string      `json:"codespace"`
+	} `json:"response"`
+}
+
+const priv = "a96e62ed3955e65be3aaa3f12d87b6b5cf26039ecfa948dc5107a495418e5430"
+const port = "5001"
+const nodeURI = "http://localhost:26657"
+const queryURI = "http://localhost:1317"
+
```
sorawit, line 84: Comment what this function does. Also should we add unit tests for this function?

chain/cmd/bandsv/main.go
```diff
@@ -18,7 +30,130 @@ type OracleRequestResp struct {
...
+func GetTx(txHash string) (TxResult, error) {
```
sorawit: Should this function accept `[]byte`. I think that makes more sense. Also prevent the case where people get confused whether they need to prefix "0x" to txHash.

chain/cmd/bandsv/main.go
```diff
@@ -18,7 +30,130 @@ type OracleRequestResp struct {
...
+		return TxResult{}, err
+	}
+
+	var response map[string]interface{}
+	if err := json.Unmarshal(responseBytes, &response); err != nil {
+		return TxResult{}, err
+	}
+
+	var txr TxResult
+	if txData, ok := response["result"]; ok {
```
sorawit: If we intend to do polling? Why don't we do it in this function?

chain/cmd/bandsv/main.go
```diff
@@ -18,7 +30,130 @@ type OracleRequestResp struct {
 	Params    []byte `json:"params"`
 }
 
+type TxResult struct {
```
sorawit: Can't we reuse data structure from Tendermint, instead of re-defining one ourselves?

chain/cmd/bandsv/main.go
```diff
@@ -18,7 +30,130 @@ type OracleRequestResp struct {
...
+	}
+	defer _resp.Body.Close()
+
+	responseBytes, err := ioutil.ReadAll(_resp.Body)
+	if err != nil {
+		return TxResult{}, err
+	}
+
+	var response map[string]interface{}
+	if err := json.Unmarshal(responseBytes, &response); err != nil {
```
sorawit: Just to confirm. Are you sure we are supposed to use `json.Unmarshal` here (and not some kind of amino encode)?

chain/cmd/bandsv/main.go
```diff
@@ -18,7 +30,130 @@ type OracleRequestResp struct {
...
+		return TxResult{}, err
+	}
+
+	var txr TxResult
```
sorawit: ame this `txResult` or `result`

chain/cmd/bandsv/main.go
sorawit: Instead of using `http.Get` and need to raw stuff like `ioutil.ReadAll`, let's use convenient library like grequests? https://github.com/levigross/grequests

chain/cmd/bandsv/main.go
```diff
@@ -18,7 +30,130 @@ type OracleRequestResp struct {
...
+		return TxResult{}, fmt.Errorf("%v", response)
+	}
+
+	return txr, nil
+}
+
+func HasCode(codeHash []byte) bool {
+	_resp, err := http.Get(
+		fmt.Sprintf(
+			`%s/abci_query?path="/store/zoracle/key"&data=0x02%016x`,
```
sorawit: Let's use function that generates key that we write in `/x/zoracle`

chain/cmd/bandsv/main.go
```diff
@@ -18,7 +30,130 @@ type OracleRequestResp struct {
...
+		}
+	} else {
+		return TxResult{}, fmt.Errorf("%v", response)
+	}
+
+	return txr, nil
+}
+
+func HasCode(codeHash []byte) bool {
+	_resp, err := http.Get(
```
sorawit: Again let's use `grequests`

chain/cmd/bandsv/main.go
```diff
@@ -35,18 +170,73 @@ func main() {
...
 		if len(json.Code) > 0 {
-			// TODO:
-			//   1. Compute code hash.
-			//   2. Check if code hash already exist on chain. If not deploy it.
-			//   3. Save code hash to `json`
-		}
-		// TODO:
-		//   1. Send the request to the blockchain
-		//   2. Wait for the tx to confirm and get back the request ID
+			codeHash = crypto.Sha256(json.Code)
```
sorawit: Instead of having this variable, I think we can mutate code hash directly.

sorawit (changes requested):

chain/cmd/bandsv/main.go
sorawit: In fact, instead of using `http.Get` or grequests, can't we use Tendermint's rpc client directly? See https://github.com/tendermint/tendermint/blob/master/lite2/rpc/client.go for example. It really seems like all the code is written for us already.

chain/cmd/bandsv/main.go
```diff
@@ -35,18 +170,73 @@ func main() {
...
+				}
+			}
+		}
+
 		c.JSON(200, OracleRequestResp{
-			RequestId: 10,
-			CodeHash:  nil,
+			RequestId: requestId,
+			CodeHash:  codeHash,
 			Params:    nil,
```
sorawit: Should we return the parameters back here?

sorawit (changes requested):

chain/cmd/bandsv/main.go
```diff
@@ -18,7 +30,130 @@ type OracleRequestResp struct {
...
+			return TxResult{}, err
+		}
+	} else {
+		return TxResult{}, fmt.Errorf("%v", response)
+	}
+
+	return txr, nil
+}
+
+func HasCode(codeHash []byte) bool {
```
taobun: If you want to check code hash already in chain, I think create new issue to add query from rest client and ask via that endpoint (Code will more robust it chain has change.)

chain/cmd/bandsv/main.go
```diff
@@ -35,18 +170,73 @@ func main() {
 			c.JSON(http.StatusBadRequest, gin.H{"error": "Only one of code/codeHash can be sent"})
```
taobun: Move handler to outer main function (write handler outside main)

chain/cmd/bandsv/main.go
```diff
@@ -35,18 +170,73 @@ func main() {
...
-			//   3. Save code hash to `json`
-		}
-		// TODO:
-		//   1. Send the request to the blockchain
-		//   2. Wait for the tx to confirm and get back the request ID
+			codeHash = crypto.Sha256(json.Code)
+			hasCode := HasCode(codeHash)
+			// If codeHash not found then store the code
+			if !hasCode {
+				tx := cmtx.NewTxSender(pk)
```
taobun: Change `tx` to `txSender` make more sense.

chain/cmd/bandsv/main.go
```diff
@@ -35,18 +170,73 @@ func main() {
 			c.JSON(http.StatusBadRequest, gin.H{"error": "Only one of code/codeHash can be sent"})
 			return
 		}
+
 		if len(json.Code) > 0 {
```
taobun: Change variable name from `json` to `requestData`

chain/cmd/bandsv/main.go
```diff
@@ -35,18 +170,73 @@ func main() {
...
+		events := txr.Result.Events
+		for _, event := range events {
+			if event.Type == "request" {
+				for _, attr := range event.Attributes {
+					k, _ := base64.StdEncoding.DecodeString(attr.Key)
+					if string(k) == "id" {
+						v, _ := base64.StdEncoding.DecodeString(attr.Value)
+						requestId, err = strconv.ParseUint(string(v), 10, 64)
+						if err != nil {
+							c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
```
taobun: I think we should return 500 response instead 400

chain/cmd/bandsv/main.go
```diff
@@ -35,18 +170,73 @@ func main() {
...
+		}
+
+		tx := cmtx.NewTxSender(pk)
+		txHash, err := tx.SendTransaction(zoracle.NewMsgRequest(codeHash, 5, tx.Sender()))
+		if err != nil {
+			c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+			return
+		}
+		// waiting for tx confirmation
+		var txr TxResult
```
taobun: TODO: Not use while true, set some limit round to loop

taobun (changes requested):

chain/cmd/bandsv/main.go
```diff
@@ -18,7 +28,45 @@ type OracleRequestResp struct {
...
 }
 
+const priv = "a96e62ed3955e65be3aaa3f12d87b6b5cf26039ecfa948dc5107a495418e5430"
+const port = "5001"
+const nodeURI = "http://localhost:26657"
+const queryURI = "http://localhost:1317"
+
+var rpcClient *rpc.HTTP
+
+func GetTx(txHash []byte) *ctypes.ResultTx {
```
sorawit: Shouldn't this return `(*ctypes.ResultTx, error)`?

chain/cmd/bandsv/main.go
sorawit: Let's leave that as a TODO. That's out of the scope of this work.

sorawit (changes requested):

sorawit:
Also there's a merge conflict. Please rebase before the next review.

chain/cmd/bandsv/main.go
```diff
@@ -70,42 +77,114 @@ type Proof struct {
...
-		}
-		if len(json.CodeHash) == 0 && len(json.Code) == 0 {
-			c.JSON(http.StatusBadRequest, gin.H{"error": "Missing code/codeHash"})
-			return
+var rpcClient *rpc.HTTP
+var pk secp256k1.PrivKeySecp256k1
+
+// TODO
+// - Add query from rest client and ask via that endpoint
+func HasCode(codeHash []byte) bool {
```
sorawit: This should return `(bool, error)`. Don't mix up error with code not exist.

chain/cmd/bandsv/main.go
```diff
@@ -70,42 +77,114 @@ type Proof struct {
...
+
+	if len(requestData.Code) > 0 {
+		requestData.CodeHash = crypto.Sha256(requestData.Code)
+		hasCode := HasCode(requestData.CodeHash)
+		// If codeHash not found then store the code
+		if !hasCode {
+			txSender := cmtx.NewTxSender(pk)
+			_, err := txSender.SendTransaction(zoracle.NewMsgStoreCode(requestData.Code, txSender.Sender()), flags.BroadcastBlock)
+			if err != nil {
+				c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
```
sorawit: 500

chain/go.mod
```diff
@@ -4,6 +4,7 @@ go 1.13
 
 require (
 	github.com/cosmos/cosmos-sdk v0.37.4
+	github.com/davecgh/go-spew v1.1.1
```
sorawit: remove

sorawit (changes requested):

chain/cmd/bandsv/main.go
```diff
@@ -70,42 +77,122 @@ type Proof struct {
...
+			}
 		}
-		// TODO:
-		//   1. Send the request to the blockchain
-		//   2. Wait for the tx to confirm and get back the request ID
-		c.JSON(200, OracleRequestResp{
-			RequestId: 10,
-			CodeHash:  nil,
-		})
+	}
```
sorawit, line 171: If we somehow reach this but `requestId` is still `0`, we should 500 to user.

sorawit (changes requested):

sorawit (approved):

sorawit (approved):

chain/cmd/bandsv/main.go
```diff
@@ -70,42 +78,126 @@ type Proof struct {
...
+
+	// TODO
+	// Need some work around to make params can be empty bytes
+	if len(requestData.Params) <= 0 {
+		c.JSON(http.StatusBadRequest, gin.H{"error": "Params should not be empty bytes"})
+		return
+	}
+
+	if len(requestData.Code) > 0 {
+		requestData.CodeHash = crypto.Sha256(requestData.Code)
```
taobun: Use getCodeHash from x/zoracle from this [file](https://github.com/bandprotocol/d3n/blob/master/chain/x/zoracle/internal/types/code.go)

chain/cmd/bandsv/main.go
```diff
@@ -70,42 +78,126 @@ type Proof struct {
...
+	txr, err := txSender.SendTransaction(zoracle.NewMsgRequest(requestData.CodeHash, requestData.Params, 4, txSender.Sender()), flags.BroadcastBlock)
+	if err != nil {
+		c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
+		return
+	}
+
+	requestId := uint64(0)
+	events := txr.Events
+	for _, event := range events {
+		if event.Type == "request" {
```
taobun, line 159: Break loop if already find requestId

docker-compose.yaml
```diff
@@ -12,8 +12,6 @@ services:
     environment:
       - PRIVATE_KEY=06be35b56b048c5a6810a47e2ef612eaed735ccb0d7ea4fc409f23f1d1a16e0b
       - NODE_URI=tcp://localhost:26657
-    volumes:
```
taobun: Ask @evilpeach how to solve this?

taobun (changes requested):

taobun (approved):

PR #100, opened 2019-12-24
Add mint module to inflate every begin block

### LINKED ISSUE #12: Add minting module
Introduce the inflation of BANDs. Adding minting features that will inflate BANDs to blockchain validators for every block produced, using the same parameters as the Cosmos chain.

taobun:
fixed #12

- Add mint module to inflate every begin block
- Inflated token send to fee collector and will be send to distributor to distribute to validators

### DIFF
--- a/chain/docker-config/single-validator/genesis.json
+++ b/chain/docker-config/single-validator/genesis.json
@@ -1,5 +1,5 @@
 {
-  "genesis_time": "2019-12-18T09:29:37.567985Z",
+  "genesis_time": "2019-12-24T04:31:10.609444Z",
   "chain_id": "bandchain",
   "consensus_params": {
     "block": {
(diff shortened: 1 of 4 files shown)


taobun:
After run chain investigate balance of distribution module by this [endpoint](http://localhost:1317/bank/balances/band1jv65s3grqf6v6jl3dp4t6c9t9rk99cd87sy73h)

chain/docker-config/single-validator/genesis.json
```diff
@@ -1,5 +1,5 @@
 {
-  "genesis_time": "2019-12-18T09:29:37.567985Z",
+  "genesis_time": "2019-12-24T04:31:10.609444Z",
```
sorawit, line 2: Please generate genesis for multi-val too

sorawit (changes requested):

sorawit (approved):

PR #104, opened 2019-12-25
Implement dedicated storage key for data result

### LINKED ISSUE #87: Implement dedicated storage key for data result
Let's make the prefix `0xff` following by Request ID, CodeHash, and Parameters.

taobun:
fixed #87 

- Don't change how to request data use go script and endpoint followed old pattern
-  After run docker and request has been executed, see result at http://localhost:1317/zoracle/request/1


chain/x/zoracle/internal/types/error.go
```diff
@@ -22,6 +23,10 @@ func ErrRequestNotFound(codespace sdk.CodespaceType) sdk.Error {
 	return sdk.NewError(codespace, CodeRequestNotFound, "request not found")
 }
 
+func ErrResultNotFound(codespace sdk.CodespaceType) sdk.Error {
+	return sdk.NewError(codespace, CodeResultNotFound, "request not found")
```
sorawit: result not found?

chain/x/zoracle/internal/keeper/result.go
```diff
@@ -0,0 +1,29 @@
...
+	store := ctx.KVStore(k.storeKey)
+	store.Set(
+		types.ResultStoreKey(requestID, codeHash, params),
+		result,
+	)
+}
+
+func (k Keeper) GetResult(ctx sdk.Context, requestID uint64, codeHash []byte, params []byte) ([]byte, sdk.Error) {
+	if !k.HasResult(ctx, requestID, codeHash, params) {
+		return []byte(nil), types.ErrResultNotFound(types.DefaultCodespace)
```
sorawit: Can't we just `return nil, ...`?

sorawit (changes requested):
Looks great. Just a few comments.

sorawit (approved):

PR #142, opened 2020-01-08
scan: fix key warning on footer

evilpeach:
- change `Belt_Array` to `Belt.Array`
- fix key warning of rendering an array


sorawit (approved):
Thanks

PR #146, opened 2020-01-08
 Implement `u64` array encoding on Owasm

### LINKED ISSUE #15: Implement `i64` array encoding on Owasm
We updated the way an array can be encoded by using just one 64 bit integer (top 32 encoding the size, and bottom 32 encoding the location of data). Let's update `owasm` code base to work like that.

### LINKED ISSUE #16: Update chain code to use `i64` array representation

taobun:
fixed #15,
fixed #16


chain/wasm/wasm.go
```diff
@@ -1,84 +1,82 @@
...
 	}
 	return res, nil
 }
 
-func storeParams(instance wasm.Instance, params []byte) (int32, error) {
+func storeParams(instance wasm.Instance, params []byte) (int64, error) {
 	return allocateInner(instance, params)
 }
 
+func Name(code []byte) ([]byte, error) {
```
sorawit: return string. error if string is shit.

chain/wasm/wasm.go
```diff
@@ -1,84 +1,82 @@
...
 	res := make([]byte, sz)
 	for idx := 0; idx < sz; idx++ {
-		res[idx] = mem[4+idx]
+		res[idx] = mem[idx]
 	}
 	return res, nil
 }
 
-func storeParams(instance wasm.Instance, params []byte) (int32, error) {
+func storeParams(instance wasm.Instance, params []byte) (int64, error) {
```
sorawit, line 60: remove

sorawit (changes requested):

sorawit (approved):
Looks good.

PR #147, opened 2020-01-08
scan: fix spacing & text color

evilpeach:
image


scan/src/Colors.re
```diff
@@ -13,4 +13,8 @@ let orange = Css.hex("FF9C9C");
 
 let green = Css.hex("6BCA4A");
 
+let greenDark = Css.hex("127658");
```
sorawit: darkGreen

sorawit (changes requested):

sorawit (approved):
Thanks!

PR #178, opened 2020-01-10
Implement endpoint to get data script information

### LINKED ISSUE #155: Implement endpoint to get data script information
Given CodeHash it should return the following information
- Script name
- List of parameters & types
- List of data sources & types
- Creator
- Created at
- Result

taobun:
fixed #155 

New endpoint found at [link](http://localhost:1317/zoracle/script/33cefd052eb9b0cda3d38d4d87313295e45fda5c5b0d6ab9e54870866f62fc80)

Example output
```
{
  "height": "0",
  "result": {
    "info": {
      "name": "Crypto price",
      "params": [
        {
          "name": "symbol_cg",
          "type": "String"
        },
        {
          "name": "symbol_cc",
          "type": "String"
        }
      ],
      "dataSources": [
        {
          "name": "coin_gecko",
          "type": "f32"
        },
        {
          "name": "crypto_compare",
          "type": "f32"
        }
      ],
      "creator": "band1r8u4m0wefrdyp7sdde7drtrt94qczggkv38jrk"
    },
    "txhash": "8375814D9439A1FB85605DD1C649F6411EC90CF1003CA086FE1AE60B8FABEA13",
    "createdAtHeight": "15",
    "createdAtTime": "2020-01-10T12:49:15Z"
  }
}
```

### DIFF
--- a/chain/x/zoracle/client/rest/query.go
+++ b/chain/x/zoracle/client/rest/query.go
@@ -7,8 +7,10 @@ import (
 	"github.com/cosmos/cosmos-sdk/client/context"
 
 	"github.com/cosmos/cosmos-sdk/types/rest"
-
+	"github.com/cosmos/cosmos-sdk/x/auth/client/utils"
 	"github.com/gorilla/mux"
+
+	"github.com/bandprotocol/d3n/chain/x/zoracle/internal/types"
 )
 
 func getRequestHandler(cliCtx context.CLIContext, storeName string) http.HandlerFunc {
@@ -29,14 +31,32 @@ func getRequestHandler(cliCtx context.CLIContext, storeName string) http.Handler
 func getScriptHandler(cliCtx context.CLIContext, storeName string) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		vars := mux.Vars(r)
-		paramType := vars[codeHash]
+		codeHash := vars[codeHash]
 
-		res, _, err := cliCtx.QueryWithData(fmt.Sprintf("custom/%s/script/%s", storeName, paramType), nil)
+		res, _, err := cliCtx.QueryWithData(fmt.Sprintf("custom/%s/script/%s", storeName, codeHash), nil)
 		if err != nil {
 			rest.WriteErrorResponse(w, http.StatusNotFound, err.Error())
 			return
 		}
 
-		rest.PostProcessResponse(w, cliCtx, res)
+		var scriptInfo types.ScriptInfo
+		err = cliCtx.Codec.UnmarshalJSON(res, &scriptInfo)
+		if err != nil {
+			rest.WriteErrorResponse(w, http.StatusNotFound, err.Error())
+			return
+		}
+
+		// Find TxHash and height that of transaction
+		// TODO: Get latest store tx as tx hash (wait tendermint release get result in desc order)
+		searchResult, err := utils.QueryTxsByEvents(
+			cliCtx,
+			[]string{fmt.Sprintf("store_code.codehash='%s'", codeHash)},
+			1,
+			1,
+		)
+		scriptInfo.TxHash = searchResult.Txs[0].TxHash
+		scriptInfo.CreatedAt = searchResult.Txs[0].Height
+
+		rest.PostProcessResponse(w, cliCtx, scriptInfo)
 	}
 }
--- a/chain/x/zoracle/client/rest/query.go
+++ b/chain/x/zoracle/client/rest/query.go
@@ -5,10 +5,11 @@ import (
 	"net/http"
 
 	"github.com/cosmos/cosmos-sdk/client/context"
-
 	"github.com/cosmos/cosmos-sdk/types/rest"
 	"github.com/cosmos/cosmos-sdk/x/auth/client/utils"
 	"github.com/gorilla/mux"
+
+	"github.com/bandprotocol/d3n/chain/x/zoracle/internal/types"
 )
 
 func getRequestHandler(cliCtx context.CLIContext, storeName string) http.HandlerFunc {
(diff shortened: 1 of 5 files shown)


chain/x/zoracle/internal/keeper/querier.go
```diff
@@ -59,3 +63,51 @@ func queryPending(ctx sdk.Context, path []string, req abci.RequestQuery, keeper
 	}
 	return res, nil
 }
+
+// queryPending is a query function to get the list of request IDs that are still on pending status.
```
sorawit: Update comment

chain/x/zoracle/internal/keeper/querier.go
```diff
@@ -59,3 +63,51 @@ func queryPending(ctx sdk.Context, path []string, req abci.RequestQuery, keeper
...
+func queryScript(ctx sdk.Context, path []string, req abci.RequestQuery, keeper Keeper) ([]byte, sdk.Error) {
+	if len(path) == 0 {
+		return nil, sdk.ErrInternal("must specify the requestid")
```
sorawit: Update error

chain/x/zoracle/internal/types/querier.go
```diff
@@ -54,3 +57,50 @@ func (rb RawBytes) String() string {
...
+}
+
+func ParseFields(raw []byte) ([]Field, error) {
+	var data [][]string
+	err := json.Unmarshal(raw, &data)
+	if err != nil {
+		return nil, err
+	}
+	fields := make([]Field, len(data))
+	for idx, row := range data {
```
sorawit, line 73: Let's check for row's size here, so the code doesn't panic when a user sends bad code to us.

chain/x/zoracle/internal/keeper/querier.go
```diff
@@ -59,3 +63,51 @@ func queryPending(ctx sdk.Context, path []string, req abci.RequestQuery, keeper
...
+	if err != nil {
+		// TODO: Return err
+		name = ""
+	}
+
+	// Get raw params info
+	rawParamsInfo, err := wasm.ParamsInfo(code.Code)
+	var paramsInfo []types.Field
+	if err != nil {
+		paramsInfo = []types.Field{}
```
sorawit: What does this code do? `paramsInfo` will get set again below.

chain/x/zoracle/internal/types/querier.go
```diff
@@ -54,3 +57,50 @@ func (rb RawBytes) String() string {
...
+	}
+	return fields
+}
+
+type ScriptInfo struct {
+	Name        string         `json:"name"`
+	Params      []Field        `json:"params"`
+	DataSources []Field        `json:"dataSources"`
+	Creator     sdk.AccAddress `json:"creator"`
+	TxHash      string         `json:"txhash"`
```
sorawit: This is confusing. TxHash/CreatedAt is not part of `NewScriptInfo`. Maybe it shouldn't be in here? I guess we can have a separate struct in client/rest/query.go

chain/x/zoracle/client/rest/query.go
```diff
@@ -25,3 +27,36 @@ func getRequestHandler(cliCtx context.CLIContext, storeName string) http.Handler
...
+		// Find TxHash and height that of transaction
+		// TODO: Get latest store tx as tx hash (wait tendermint release get result in desc order)
+		searchResult, err := utils.QueryTxsByEvents(
+			cliCtx,
+			[]string{fmt.Sprintf("store_code.codehash='%s'", codeHash)},
+			1,
+			1,
+		)
+		scriptInfo.TxHash = searchResult.Txs[0].TxHash
+		scriptInfo.CreatedAt = searchResult.Txs[0].Height
```
sorawit: Let's have `CreatedAtTime` for timestamp. `CreatedAtHeight` for block height.

chain/x/zoracle/client/rest/query.go
```diff
@@ -25,3 +27,36 @@ func getRequestHandler(cliCtx context.CLIContext, storeName string) http.Handler
...
+		if err != nil {
+			rest.WriteErrorResponse(w, http.StatusNotFound, err.Error())
+			return
+		}
+
+		// Find TxHash and height that of transaction
+		// TODO: Get latest store tx as tx hash (wait tendermint release get result in desc order)
+		searchResult, err := utils.QueryTxsByEvents(
+			cliCtx,
+			[]string{fmt.Sprintf("store_code.codehash='%s'", codeHash)},
```
sorawit: Is there a way we can avoid hard-coding the string by reusing the ones we defined somewhere else?

chain/x/zoracle/client/rest/query.go
```diff
@@ -25,3 +27,36 @@ func getRequestHandler(cliCtx context.CLIContext, storeName string) http.Handler
 		rest.PostProcessResponse(w, cliCtx, res)
 	}
 }
+
+func getScriptHandler(cliCtx context.CLIContext, storeName string) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		vars := mux.Vars(r)
+		codeHash := vars[codeHash]
```
sorawit: This will never fail, right?

chain/x/zoracle/client/rest/query.go
```diff
@@ -25,3 +27,36 @@ func getRequestHandler(cliCtx context.CLIContext, storeName string) http.Handler
...
+
+		var scriptInfo types.ScriptInfo
+		err = cliCtx.Codec.UnmarshalJSON(res, &scriptInfo)
+		if err != nil {
+			rest.WriteErrorResponse(w, http.StatusNotFound, err.Error())
+			return
+		}
+
+		// Find TxHash and height that of transaction
+		// TODO: Get latest store tx as tx hash (wait tendermint release get result in desc order)
```
sorawit, line 49: I'm not sure I understand this TODO. We are querying for only one transaction?

sorawit (changes requested):

chain/x/zoracle/client/rest/query.go
taobun: In case there are tx that store the same codehash

chain/x/zoracle/client/rest/query.go
taobun: How to fail?

sorawit: There's a bit of variable shadowing in here, right? the right codeHash is the global variable, while the left is a new variable. Is this intended?

What happens if user does not specify codeHash? I guess it will try to query with empty string and fail on the line after this, correct?

chain/x/zoracle/client/rest/query.go
sorawit: Make sense. In the future when a code is added, then deleted, then added again.

chain/x/zoracle/client/rest/query.go
taobun: 1. Good catch change name already
2. If not specify, route will be incorrect and not execute handler function (404 return)

sorawit (approved):
LGTM Thanks!

PR #187, opened 2020-01-12
Implement hook to get code script information by code hash

### LINKED ISSUE #180: Implement hook to get code script information by code hash

taobun:
fixed #180

### DIFF
--- a/scan/src/rpc/ScriptHook.re
+++ b/scan/src/rpc/ScriptHook.re
@@ -1,5 +1,5 @@
 module Script = {
-  type field = {
+  type field_t = {
     name: string,
     dataType: string,
   };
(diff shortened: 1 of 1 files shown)


scan/src/rpc/ScriptHook.re
```diff
@@ -0,0 +1,41 @@
+module Script = {
+  type field = {
```
sorawit: Let's name this `field_t`. It's a bit confusing that we declare `field` name here and also `field`  from `JsonUtils.Decode` below.

sorawit (changes requested):
Looks good. Just a small comment.

sorawit (approved):
Looks great!

PR #221, opened 2020-01-14
Emit script name during zoracle transactions

### LINKED ISSUE #166: Emit script name during zoracle transactions

taobun:
fixed #166 

- Every msg will emit code name in event
- Code name must send to chain via MsgStoreCode
- Deprecate `Name()` in wasm part

### DIFF
--- a/chain/cmd/bandsv/main.go
+++ b/chain/cmd/bandsv/main.go
@@ -32,6 +32,7 @@ import (
 type OracleRequest struct {
 	CodeHash []byte `json:"codeHash" binding:"len=0|len=32"`
 	Code     []byte `json:"code"`
+	Name     string `json:"name"`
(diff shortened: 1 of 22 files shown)


sorawit (approved):
@taobun  LGTM thanks for getting this up quickly!
@prin-r  Please review this change. After this is landed, you will need to update `scan` code to reflect this change as part of #189

sorawit (changes requested):
Actually, don't we need to remove code that performs `__name`? Let's do that both in `wasm.go` and `lib.rs`.

sorawit (approved):
LGTM. @prin-r This is how you will get the name of code script for description of messages. Please review.

prin-r (approved):

PR #224, opened 2020-01-14
Add temporary implementation to make /d3n/txs/latest perform correctly

### LINKED ISSUE #191: Add temporary implementation to make /d3n/txs/latest perform correctly
While we are waiting for tendermint/tendermint#4253 to get merged and released, let's implement a hacky version first. The code should just query for ALL transactions (page=1 limit=1,000,000), reverse the list, and return the correct data at the requested page / limit.

taobun:
fixed #191 

Use math to calculate index of transaction in page

### DIFF
--- a/chain/client/client.go
+++ b/chain/client/client.go
@@ -76,7 +76,8 @@ func LatestTxsRequestHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
 		// TODO: (1) Sort result in desc order after tendermint/tendermint:#4253 is released
 		// TODO: (2) Perform binary search on 'tx.height>?' to optimize the performance
 
-		// Get all latest txs
+		// Temporary implementation to get latest tx sort by descending timestamp
+		// Pull request at bandprotocol/d3n:#224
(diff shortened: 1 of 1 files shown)


chain/client/client.go
```diff
@@ -74,13 +75,30 @@ func LatestTxsRequestHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
 
 		// TODO: (1) Sort result in desc order after tendermint/tendermint:#4253 is released
 		// TODO: (2) Perform binary search on 'tx.height>?' to optimize the performance
-		searchResult, err := utils.QueryTxsByEvents(cliCtx, []string{"tx.height>0"}, page, limit)
+
+		// Get all latest txs
```
sorawit: Add comment why we do this. Also reference to this PR number so we can go back here.

sorawit (changes requested):

chain/client/client.go
sorawit: @prin-r After this is landed and deployed. You can happily do wireup all txs page.

sorawit (approved):
lgtm thanks

prin-r (approved):

PR #229, opened 2020-01-15
Move get proof API to be part of rest-server endpoint

### LINKED ISSUE #172: Move get proof API to be part of rest-server endpoint

taobun:
fixed #172 

This pull request just move and modify code from `bandsv/main.go` to `client/proof.go`. Make sure get the same result with old endpoint and can report to chain in functions `relayOracleState` and `verifyOracleData`

* This pull doesn't include generating 1 big bytes proof used in `relayAndVerify`, we will open new issue to make sure that function work and add field to this endpoint later

### DIFF
--- a/chain/client/client.go
+++ b/chain/client/client.go
@@ -1,6 +1,7 @@
 package rpc
 
 import (
+	"fmt"
 	"net/http"
 
 	"github.com/gorilla/mux"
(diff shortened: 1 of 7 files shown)


sorawit (approved):

sorawit:
@prin-r

prin-r (approved):

PR #236, opened 2020-01-16
Fix validator report bug by forking tendermint

### LINKED ISSUE #61: [Bug] Tx subscriber `Server error: subscription was cancelled (reason: client is not pulling messages fast enough)
Event subscriber from tendermint stop subscribe after listen 4 reports in one block

How to produce?
Run chain with 4 validators when request has been executed. All 4 validators send report to chain in the same block, but tendermint server cannot handle report event correctly. (Internal buffer is full with capacity = 1)

Possible root cause
After experiment error come from `https://github.com/tendermint/tendermint/blob/master/libs/pubsub/pubsub.go` in `send` function. Code cannot push new event to channel because it's full. Consumer is `https://github.com/tendermint/tendermint/blob/master/rpc/core/events.go` that consume data after new event push make pubsub return error and try to cancel subscription.

- But ws client have code to resub after got error but code doesn't work out as expect. It return error told that already subscribe

Way to solve
- Make sure consumer consume event fast enough (?) for new event
- Increase buffer size of buffer
- Make resubscription work as code expected

taobun:
fixed #61 

- Using our tendermint fork as dependency at this [commit](https://github.com/bandprotocol/tendermint/tree/modify-event-code)
- Add new role `requester` for store and request data instead of validator account

### DIFF
--- a/chain/docker-config/genesis.json
+++ b/chain/docker-config/genesis.json
@@ -1,5 +1,5 @@
 {
-  "genesis_time": "2020-01-03T05:36:57.439728Z",
+  "genesis_time": "2020-01-15T07:59:16.798155Z",
   "chain_id": "bandchain",
   "consensus_params": {
     "block": {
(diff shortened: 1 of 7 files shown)


chain/docker-config/genesis.json
```diff
@@ -1,5 +1,5 @@
 {
-  "genesis_time": "2020-01-03T05:36:57.439728Z",
+  "genesis_time": "2020-01-15T07:59:16.798155Z",
```
sorawit, line 2: Please also update single validator genesis

sorawit (changes requested):

sorawit (approved):

evilpeach (approved):

PR #256, opened 2020-01-16
Make get proof data endpoint return hex data to send to contract

### LINKED ISSUE #231: Make get proof data endpoint return hex data to send to Bridge's `relayAndVerify`
Note that `relayAndVerify` is not tested yet. Verifying that the function works is also part of this ticket.

taobun:
fixed #231 

- Fixed how to encode parameter to call other function in contract.
- Copy ABI to string in go client to calculate single byte proof
- Use abi package from `github.com/ethereum/go-ethereum/accounts/abi`


chain/client/proof.go
```diff
@@ -417,10 +520,34 @@ func GetProofHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
...
+		proofBytes, err := relayAndVerifyArguments.Pack(blockRelayBytes, oracleDataBytes)
+		if err != nil {
+			rest.WriteErrorResponse(w, http.StatusInternalServerError, err.Error())
+		}
+
 		rest.PostProcessResponse(w, cliCtx, Proof{
 			BlockHeight:     height,
 			OracleDataProof: odp,
 			BlockRelayProof: brp,
+			ProofBytes:      proofBytes,
```
sorawit: EVMProofBytes

sorawit (changes requested):

sorawit (approved):
LGTM. @prin-r

prin-r (approved):

PR #260, opened 2020-01-16
Implement proof of validity and copy proof for Ethereum

### LINKED ISSUE #199: Implement "Proof of Validity" tab

### LINKED ISSUE #200: Implement "Copy proof for Ethereum"
It should copy to clipboard bytes data that can be use as the argument for `relayAndVerify` function in Bridge.sol

prin-r:
Fixes #199 
Fixes #200


scan/src/components/ProofTab.re
```diff
@@ -0,0 +1,60 @@
+let copyStringToClipboard: (string, string) => unit = [%bs.raw
+  {|
+function copyStringToClipboard (str,strMsg) {
+   var element = document.createElement('textarea');
```
sorawit: What's the difference between using this vs using https://www.npmjs.com/package/clipboard-copy, similar to what we did in staking portal. 

Can you confirm the solution you are doing will work on all platforms?

sorawit (changes requested):

scan/src/components/ProofTab.re
```diff
@@ -0,0 +1,66 @@
...
+let make = (~reqID) => {
+  let proofOpt = ProofHook.get(~requestId=reqID, ());
+
+  let (innerElement, evmProofHex) =
+    switch (proofOpt) {
+    | Some({jsonProof, evmProofBytes}) => (
+        <ReactHighlight>
+          {jsonProof |> Js.Json.stringifyWithSpace(_, 2) |> React.string}
+        </ReactHighlight>,
+        evmProofBytes |> JsBuffer.toHex,
```
sorawit: Why don't you send the `0x`  prefixed version here, so you don't need to do that conversion again.

sorawit (changes requested):

sorawit (approved):
lgtm

evilpeach (approved):

PR #274, opened 2020-01-20
scan: update values and params to use Js.Json.t

### LINKED ISSUE #267: Make `values` of Report.t and `params` of Request.t be `array((string, Js.Json.t))`
We implemented a quick hack to use `float` and `string` type respectively, but it fact the values can be any JSON type. Let's fix that. To show the value, we can use `Js.Json.stringify` for now, which can convert any `Js.Json.t` to `string`.

evilpeach:
Fixes #267


sorawit (approved):
LGTM

prin-r (approved):

PR #275, opened 2020-01-20
Set up code execution timeout, both for prepare and execution

### LINKED ISSUE #137: Set up code execution timeout, both for prepare and execution
This is a temporary solution before metering is implemented. We want to limit the time it take to run code to 100ms to prevent bad infinite loop code. Note this is non-deterministic and will need to be fixed with metering.

prin-r:
Fixes #137


chain/wasm/wasm.go
```diff
@@ -158,6 +160,31 @@ func Prepare(code []byte, params []byte) ([]byte, error) {
...
+	select {
+	case <-time.After(limitTimeout):
+		return wasm.Value{}, fmt.Errorf("wasm execution timeout")
+	case res = <-chanWasmOutput:
+		if res.err != nil {
+			return wasm.Value{}, res.err
+		}
+	}
+
+	return res.ptr, res.err
```
sorawit: Do we need to kill the wasm process, or does everything get killed automatically?

sorawit (changes requested):
@prin-r Change on prepare function?

chain/wasm/wasm.go
```diff
@@ -176,7 +203,7 @@ func Execute(code []byte, params []byte, inputs [][]byte) ([]byte, error) {
 	if fn == nil {
 		return nil, errors.New("__execute not implemented")
 	}
-	ptr, err := fn(paramsInput, wasmInput)
+	ptr, err := executeWithTimeout(fn, paramsInput, wasmInput, 100*time.Millisecond)
```
taobun: 0.1 sec is too fast. I think 1 seconds still fine

taobun (changes requested):

chain/wasm/wasm.go
```diff
@@ -158,6 +160,31 @@ func Prepare(code []byte, params []byte) ([]byte, error) {
...
+func executeWithTimeout(fn func(...interface{}) (wasm.Value, error), paramsInput int64, wasmInput int64, limitTimeout time.Duration) (wasm.Value, error) {
+	type wasmOutput struct {
+		ptr wasm.Value
+		err error
+	}
+	chanWasmOutput := make(chan wasmOutput, 1)
+
+	go func() {
+		ptr, err := fn(paramsInput, wasmInput)
+		chanWasmOutput <- wasmOutput{ptr: ptr, err: err}
```
evilpeach: use `chanWasmOutput <- wasmOutput{ptr,err}` instead ?

chain/wasm/wasm.go
```diff
@@ -185,17 +174,43 @@ func Prepare(code []byte, params []byte) ([]byte, error) {
...
-	}
-	ptr, err := fn(paramsInput)
+	ptr, err := executeWithTimeout(100*time.Millisecond, instance, "__prepare", paramsInput)
 	if err != nil {
 		return nil, err
 	}
 	return parseOutput(instance, ptr.ToI64())
 }
 
+func executeWithTimeout(limitTimeout time.Duration, inst wasm.Instance, funcName string, args ...interface{}) (wasm.Value, error) {
```
sorawit: Let's pass around instance pointer: https://github.com/wasmerio/go-ext-wasm/blob/master/wasmer/instance.go#L51

sorawit (changes requested):

taobun (approved):

sorawit (approved):

PR #276, opened 2020-01-20
owasm: Make result a data structure

### LINKED ISSUE #202: owasm: Make result a data structure
Currently result is an obscure type. Let's make it inspectable just like params.
- Add `decl_result` macro. Should be really similar to `decl_params`.
- Add `__result_info` and `__parse_result`.
- Make request information endpoint also return data in the form of JSON (in addition to hex format).

taobun:
fixed #202 

- Declare new macro to specific result structure
- Export new function to get result info and parse result
- Update script, request endpoint to show result info and parsed result


owasm/src/macros.rs
```diff
@@ -11,8 +11,9 @@
 /// getting Bitcoin price from CoinGecko, and `crypto_compare` for getting Bitcoin price from CryptoCompare.
 ///
 /// ```
-/// use owasm::{decl_data, decl_params};
-/// use owasm::ext::crypto::{coingecko, cryptocompare, coins};
+/// use owasm::{decl_data, decl_params, decl_result};
+/// use owasm::ext::crypto::{coingecko, coins, cryptocompare};
+/// use serde::{Deserialize, Serialize};
```
sorawit: Why is this necessary? Can't we `use` inside of the macro?

sorawit (changes requested):
Looks good overall. I see that the CI is failing. Can you confirm why / fix it?

chain/x/zoracle/internal/types/querier.go
```diff
@@ -23,28 +23,31 @@ func (u64a U64Array) String() string {
 
 type RequestInfo struct {
 	CodeHash    []byte            `json:"codeHash"`
-	ParamsHex   []byte            `json:"paramsHex"`
 	Params      RawJson           `json:"params"`
+	ParamsRaw   []byte            `json:"paramsRaw"`
```
sorawit, line 27: @evilpeach Note this field change. Will need to update `scan` once this is live.

sorawit (approved):
LGTM. @evilpeach Please note the API change.

evilpeach (approved):

PR #291, opened 2020-01-21
Make provider run command inside of docker

### LINKED ISSUE #18: Make provider run command inside of docker
To prevent potential exploit directly to the host, all commands should be run inside of a one-time-purpose docker instance. This also helps with having one consistent docker runtime for all data providers to support similar programs with similar versions / environments.

taobun:
fixed #18 

- Add docker to `band-validator` image
- Provider run `docker run` instead of call directly as sub process


docker-provider/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM alpine:3.7
```
sorawit: `chain/images/provider-vm/Dockerfile`?

chain/Dockerfile
```diff
@@ -5,7 +5,7 @@ WORKDIR /zoracle
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
-    curl \
+    curl docker.io\
```
evilpeach: `curl docker.io \` add more space

sorawit (changes requested):

docker-provider/Dockerfile
```diff
@@ -0,0 +1,5 @@
...
+
+RUN apk update && \
+    apk add curl && \
```
evilpeach: `apk add --no-cache`

evilpeach (approved):

sorawit (commented):
LGTM. @prin-r Please review this change.

prin-r (approved):

sorawit (approved):

PR #304, opened 2020-01-23
Add support for date command in owasm prepare

### LINKED ISSUE #270: Add support for `date` command in owasm prepare
In addition to `curl`, `date` will give the data requester an access to datetime library.

prin-r:
Fixes #270


owasm/src/ext/utils/date.rs
```diff
@@ -0,0 +1,41 @@
...
+
+impl Oracle for Date {
+    type T = u64;
+
+    fn as_cmd(&self) -> ShellCmd {
+        ShellCmd::new("date", &["+'%s'"])
+    }
+
+    fn from_cmd_output(&self, output: String) -> Option<u64> {
+        output.replace("\n", "").replace("'", "").parse::<u64>().ok()
```
sorawit: Why is `replace("'", "")` necessary? What if we do `ShellCmd::new("date", &["+%s"])`?

sorawit (changes requested):

sorawit (approved):

taobun (approved):

PR #305, opened 2020-01-24
Implement driver to get yahoo finance data

### LINKED ISSUE #284: Implement driver to get yahoo finance data
Based on https://github.com/ranaroussi/yfinance, it seems you can get stock data by scraping https://finance.yahoo.com/quote/[TICKER]. Let's implement a library in owasm rust for this. Look at https://github.com/ranaroussi/yfinance/blob/master/yfinance/base.py#L414 for and https://github.com/ranaroussi/yfinance/blob/master/yfinance/utils.py#L45 for starting point.

prin-r:
Fixes #284

### DIFF
--- a/owasm/example/src/logic.rs
+++ b/owasm/example/src/logic.rs
@@ -8,21 +8,21 @@ use std::convert::TryInto;
 
 decl_params! {
     pub struct Parameter {
-        pub symbol: coins::Coins,
-        pub alphavantage_symbol: String,
+        pub crypto_symbol: coins::Coins,
+        pub stock_symbol: String,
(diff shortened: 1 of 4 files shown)


owasm/example/src/logic.rs
```diff
@@ -1,4 +1,5 @@
 use owasm::ext::crypto::{binance, coingecko, coins, cryptocompare};
+use owasm::ext::finance::yahoo_finance;
```
sorawit: Name this `owasm::ext::stock`?

taobun (approved):

sorawit (approved):

PR #306, opened 2020-01-24
scan: disable useCache on useAxios

### LINKED ISSUE #286: Hooks should be reload immediately on page change
Currently if we switch out of all txs page, then go back. We see old txs for multiple seconds. This needs to be fixed. Same for other pages that depend on hooks in the same way.

evilpeach:
Fixes #286


sorawit (approved):
Looks good.

prin-r (approved):

PR #334, opened 2020-01-28
scan: support logs optional in TxHook

evilpeach:


scan/src/rpc/TxHook.re
```diff
@@ -237,7 +237,10 @@ module Tx = {
       gasUsed: json |> field("gas_used", intstr),
       messages: {
         let actions = json |> at(["tx", "value", "msg"], list(Msg.decodeAction));
-        let eventDoubleLists = json |> field("logs", list(Event.decodeEvents));
+        let eventDoubleLists =
+          json
+          |> field("logs", optional(list(Event.decodeEvents)))
+          |> Belt.Option.getWithDefault(_, []);
```
sorawit, line 236: I'm sure this will cause failure in other places where the existence of log entries are assumed. Please make sure things won't fail

sorawit (changes requested):

sorawit (approved):

prin-r (approved):

PR #349, opened 2020-01-29
scan: use real code from server

### LINKED ISSUE #340: Implement CodeHook
Currently, CodeHook has a mocked code. Please use Axios to query real data from an endpoint.

evilpeach:
Fixes #340


scan/src/rpc/CodeHook.re
```diff
@@ -1,84 +1,27 @@
...
-    #[test]
-    fn test_end_to_end_from_local_env() {
-        // Run with local environment
-        let data = Data::build_from_local_env().unwrap();
-        println!("Current BTC price (times 100) is {}", execute(vec![data]));
-    }
-}|f},
-    },
-  ];
+let getCode = (codeHash: Hash.t) => {
```
sorawit, line 18: What happens if the code can't be found? We need to make sure we handle that gracefully.

scan/src/utils/Hash.re
```diff
@@ -9,6 +9,9 @@ let toBase64 =
   fun
   | Hash(hash) => hash->JsBuffer.hexToBase64;
 
-let toHex = (~with0x=false) =>
+let toHex = (~with0x=false, ~upper=false) =>
```
sorawit, line 12: Please add unit tests to `Hash_test.re`. Consult @prin-r if you have questions.

sorawit (changes requested):

scan/src/rpc/CodeHook.re
evilpeach: In the case of any error, the ScriptCode Component will render "Code Not Found".

sorawit (dismissed):
lgtm

prin-r (approved):

sorawit (approved):

PR #356, opened 2020-01-29
Remove test and change default logic code

taobun:
There are some bugs in execution big wasm, so I decided to change default logic code that make sure can compile test and request to chain without errors. If big wasm bug is resolved, please update this default code to new version too.


taobun:
@sorawit @prin-r please review

sorawit (approved):

prin-r (approved):

PR #363, opened 2020-01-29
Add more enums for top 10 from coinmarketcap to owasm

### LINKED ISSUE #359: Add more Enums for top 10 from CoinMarketCap to OWASM

prin-r:
Fixes #359

### DIFF
--- a/owasm/src/ext/crypto/coins.rs
+++ b/owasm/src/ext/crypto/coins.rs
@@ -2,7 +2,15 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Serialize, Deserialize, PartialEq)]
 pub enum Coins {
-    BTC,
-    ETH,
+    ADA,
     BAND,
+    BCH,
+    BNB,
+    BTC,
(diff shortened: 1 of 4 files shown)


taobun (approved):
For top 10 this code make sense, but we need to think if we support more coins but some market don't have data how we handle that case. Current code it will get None of data after parse all market, but for now lt still ok.

owasm/src/ext/crypto/coins.rs
```diff
@@ -2,7 +2,15 @@ use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Serialize, Deserialize, PartialEq)]
 pub enum Coins {
+    ADA,
```
sorawit, line 5: @taobun @prin-r Important to note that doing this breaks backward compatibility with old scripts that use older version of owasm because the enum underlying values are mixed up. 

Better way to do it (if this is already live) is to add enum values after the existing ones.

sorawit (approved):

PR #364, opened 2020-01-30
Multi backends varnish

evilpeach:
- support multiple backends
- doesn't cache bandsv

### DIFF
--- a/scan/proxy/default.vcl
+++ b/scan/proxy/default.vcl
@@ -19,7 +19,7 @@ sub vcl_deliver {
 }
 
 sub vcl_recv {
-  if (req.url == "^/bandsv/") {
+  if (req.url ~ "^/bandsv/") {
     set req.url = regsub(req.url, "^/bandsv", "/");
     set req.backend_hint = bandsv;
   } else {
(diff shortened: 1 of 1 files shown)


scan/proxy/default.vcl
```diff
@@ -5,14 +5,30 @@ backend default {
...
+  if (req.url ~ "^/bandsv/") {
+    set req.url = regsub(req.url, "^/bandsv", "/");
+    set req.backend_hint = bandsv;
+  } else {
+    set req.backend_hint = default;
+  }      
+}
+
 sub vcl_backend_response {
+  if (!(bereq.url == "^/bandsv/")) {
```
sorawit: How did we confirm this does the job we want? It's weird that the logic here is different from line 22.

sorawit (changes requested):

scan/proxy/default.vcl
evilpeach: As I tested, remove this condition that works correctly

sorawit (approved):

taobun (approved):

PR #369, opened 2020-01-30
Add top 10 coins from coinmarketcap to scan

### LINKED ISSUE #361: Add top 10 coins from CoinMarketCap to scan

prin-r:
Fixes #361

### DIFF
--- a/scan/src/components/ScriptExecute.re
+++ b/scan/src/components/ScriptExecute.re
@@ -62,7 +62,6 @@ let parameterInput = (name, dataType, value, updateData) => {
      | "coins::Coins" =>
        <div className=Styles.selectPadding>
          <select
-           placeholder="Input Parameter here"
            value
            onChange={event => {
              let newVal = ReactEvent.Form.target(event)##value;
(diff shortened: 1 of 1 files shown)


scan/src/components/ScriptExecute.re
```diff
@@ -83,17 +117,17 @@ let reducer = _state =>
 [@react.component]
 let make = (~script: ScriptHook.Script.t) => {
   let params = script.info.params;
-  let preData = params->Belt.List.map(({name}) => (name, ""));
+  let preData = params->Belt.List.map(({name, dataType}) => (name, dataType, ""));
```
sorawit, line 119: Why is data type necessary here? Don't see how it's being used.

sorawit (changes requested):

scan/src/components/ScriptExecute.re
sorawit: NVM. it's being used when creating `parameterInput`. Though I think it makes more sense to just use params directly, maybe using `Belt.List.zip`? Up to you.

sorawit (approved):

evilpeach (approved):
LGTM

PR #370, opened 2020-01-30
Example contract2: Lucky number

taobun:
- Add contract to demonstrate how to use band protocol rng
- Fix owasm studio front-end can send number(`u64`, `f64`) correctly
- Update script can receive params when execute

### DIFF
--- a/bridges/evm/contracts/example/LuckyNumber.sol
+++ b/bridges/evm/contracts/example/LuckyNumber.sol
@@ -1,0 +1,56 @@
+pragma solidity 0.5.14;
+pragma experimental ABIEncoderV2;
+
+interface Bridge {
+    struct VerifyOracleDataResult {
+        bytes data;
+        bytes32 codeHash;
+        bytes params;
+    }
+    function relayAndVerify(bytes calldata data)
+        external
+        returns (VerifyOracleDataResult memory result);
+}
+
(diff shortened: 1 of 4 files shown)


bridges/evm/contracts/example/LuckyNumber.sol
```diff
@@ -0,0 +1,56 @@
+pragma solidity 0.5.14;
+pragma experimental ABIEncoderV2;
+
+interface Bridge {
+    struct VerifyOracleDataResult {
```
sorawit, line 5: Why do we use 4 space here?

bridges/evm/contracts/example/LuckyNumber.sol
```diff
@@ -0,0 +1,56 @@
...
+        );
+
+        require(result.codeHash == codeHash, "INVALID_CODEHASH");
+
+        require(maxNumber == bytesToU64(result.params), "INVALID_MAX_NUM");
+        require(target == bytesToU64(result.data), "WRONG_GUESS");
+        msg.sender.transfer(address(this).balance);
+    }
+
+    function withdraw() public {
```
sorawit, line 53: only owner?

sorawit (changes requested):

PR #375, opened 2020-01-31
Allow getting proof with 3 precommits

sorawit:
In the case where 1 of the 4 validators on testnet stops helping produce blocks, the API should still return the proof.

### DIFF
--- a/chain/client/proof.go
+++ b/chain/client/proof.go
@@ -295,7 +295,7 @@ func GetBlockRelayProof(cliCtx context.CLIContext, blockId uint64) (BlockRelayPr
 		// 	lr1,
 		// })
 	}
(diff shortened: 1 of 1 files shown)


chain/client/proof.go
```diff
@@ -295,7 +295,7 @@ func GetBlockRelayProof(cliCtx context.CLIContext, blockId uint64) (BlockRelayPr
 		// 	lr1,
 		// })
 	}
-	if len(addrs) < 4 {
+	if len(addrs) < 3 {
```
sorawit, line 298: @evilpeach FYI why getting proof takes forever sometimes.

taobun (approved):

prin-r (approved):

PR #376, opened 2020-01-31
Hide broken nonexistent links

sorawit:


scan/src/components/Footer.re
```diff
@@ -50,17 +50,17 @@ let footerData = [|
     1.0, // flex size
     [|
       ("https://bandprotocol.com", "Band Protocol Website"),
-      ("https://google.com", "D3N Blockchain"),
-      ("https://metamask.io", "D3N Wallet"),
+      ("https://scan.d3n.bandprotocol.com", "D3N Blockchain"),
```
evilpeach: IMO, we can remove this one because it links to itself.

sorawit: Yeah agree. Thanks for the suggestion.

evilpeach (approved):

taobun (approved):
LGTM

PR #377, opened 2020-01-31
Add query node for multi-validator build

taobun:


chain/cmd/request/main.go
```diff
@@ -81,14 +81,14 @@ func main() {
 					// BTC parameter
 					params, _ := hex.DecodeString("00000000")
 					// Send request by code hash
-					fmt.Println(tx.SendTransaction(zoracle.NewMsgRequest(codeHash, params, 4, tx.Sender()), flags.BroadcastBlock))
+					fmt.Println(tx.SendTransaction(zoracle.NewMsgRequest(codeHash, params, 6, tx.Sender()), flags.BroadcastBlock))
```
sorawit, line 84: Why is this changed?

sorawit (changes requested):

chain/cmd/request/main.go
taobun: I tried to run in local mode, 4 block time duration is not enough to get data and report in time. After investigation, I found report miss in a little of time after change it to 6 blocks. All providers report successfully.

sorawit: Make sense.

sorawit (approved):

prin-r (approved):

PR #382, opened 2020-01-31
Change varnish config  (ip of rest-server)

taobun:
After change ip in docker-compose need to update config in varnish to point to new rest-server service ip fix bug after merge #377


sorawit (approved):
lgtm

PR #412, opened 2020-02-03
Make proof faster

### LINKED ISSUE #378: Make proof faster
- [x] Try bandd start pruning with options nothing and everything. It make we can get proof in current block
- [ ] Update `proof.go` to get proof in current block

prin-r:
See https://github.com/cosmos/scaffold/blob/master/templates/lvl-1/cmd/aud/main.go.tmpl#L85


sorawit (commented):
Ref https://github.com/cosmos/scaffold/blob/master/templates/lvl-1/cmd/aud/main.go.tmpl#L85

sorawit (approved):

taobun (approved):

PR #419, opened 2020-02-03
Remove waiting loop for the next block, because we are now can get state from previous block

### LINKED ISSUE #413: Update proof.go to get proof in current block

prin-r:
Fixes #413

### DIFF
--- a/chain/client/proof.go
+++ b/chain/client/proof.go
@@ -9,7 +9,6 @@ import (
 	"sort"
 	"strconv"
 	"strings"
-	"time"
 
 	"github.com/cosmos/cosmos-sdk/client/context"
 	"github.com/cosmos/cosmos-sdk/codec"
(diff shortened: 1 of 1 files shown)


sorawit (changes requested):
This is great!

sorawit (dismissed):

chain/client/proof.go
```diff
@@ -417,6 +416,14 @@ func GetProofHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
 			return
 		}
 
+		rc, err := cliCtx.Client.Commit(nil)
```
taobun, line 419: Unused err?

taobun (changes requested):

chain/client/proof.go
sorawit: Good catch!

prin-r: done

taobun (approved):
lgtm

sorawit (approved):

PR #421, opened 2020-02-04
chain: Upgrade to be compatible with gaia 2.0.4

sorawit:
Go through `app.go`, `bandd/main.go`, and `bandcli/main.go` to make sure it does pretty much the same thing as what is https://github.com/cosmos/gaia/tree/release/v2.0.4 to ensure our code won't miss anything.

### DIFF
--- a/chain/app/app.go
+++ b/chain/app/app.go
@@ -263,8 +263,7 @@ func NewBandApp(
 	)
 
 	app.mm.SetOrderBeginBlockers(mint.ModuleName, distr.ModuleName, slashing.ModuleName)
-	app.mm.SetOrderEndBlockers(zoracle.ModuleName, gov.ModuleName, staking.ModuleName)
-	app.mm.SetOrderEndBlockers(crisis.ModuleName, gov.ModuleName, staking.ModuleName)
+	app.mm.SetOrderEndBlockers(zoracle.ModuleName, crisis.ModuleName, gov.ModuleName, staking.ModuleName)
 
 	// Sets the order of Genesis - Order matters, genutil is to always come last
 	// NOTE: The genutils moodule must occur after staking so that pools are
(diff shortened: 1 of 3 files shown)


chain/app/app.go
```diff
@@ -258,6 +264,7 @@ func NewBandApp(
 
 	app.mm.SetOrderBeginBlockers(mint.ModuleName, distr.ModuleName, slashing.ModuleName)
 	app.mm.SetOrderEndBlockers(zoracle.ModuleName, gov.ModuleName, staking.ModuleName)
+	app.mm.SetOrderEndBlockers(crisis.ModuleName, gov.ModuleName, staking.ModuleName)
```
taobun: Let's merge these lines

chain/cmd/bandcli/main.go
```diff
@@ -58,10 +59,11 @@ func main() {
 		client.NewCompletionCmd(rootCmd, true),
 	)
 
-	executor := cli.PrepareMainCmd(rootCmd, "REQ", app.DefaultCLIHome)
+	executor := cli.PrepareMainCmd(rootCmd, "BAND", app.DefaultCLIHome)
```
taobun, line 62: WOW

taobun (changes requested):

chain/app/app.go
```diff
@@ -31,18 +32,18 @@ import (
 )
 
 const (
-	appName          = "band"
+	appName          = "BandApp"
```
prin-r, line 35: Not relevant to the "Upgrade to be compatible with gaia 2.0.4"
But it's ok.

prin-r (approved):

chain/app/app.go
taobun: True

taobun (approved):

PR #429, opened 2020-02-04
Implement BandProvider package

### LINKED ISSUE #392: Implement BandProvider package
Reimplementation our `TxSender` to `BandProvider` that can set config when sending transaction.

- [ ] Add option using in sending transaction

taobun:
fixed #392 

Re- implement tx sender to band provider

Now can send specific nonce in transaction (if send 0 it will ask chain for nonce)


chain/provider/provider.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func privKeyToBandAccAddress(privKey crypto.PrivKey) sdk.AccAddress {
+	config := sdk.GetConfig()
+	SetBech32AddressPrefixes(config)
+	return sdk.AccAddress(privKey.PubKey().Address().Bytes())
+}
+
+// NewBandProvider create new BandProvider create new cliCtx and txBldr
+func NewBandProvider(privKey crypto.PrivKey) BandProvider {
```
sorawit: Let's return error here instead of panic, so the caller can decide what to do.

chain/provider/provider.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"github.com/cosmos/cosmos-sdk/client/context"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/auth/client/utils"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	"github.com/tendermint/tendermint/crypto"
+)
+
+// BandProvider contain context, txBuilder, private key, and address
```
sorawit: contains

chain/provider/provider.go
```diff
@@ -0,0 +1,67 @@
...
+	config.SetBech32PrefixForConsensusNode(Bech32MainPrefix+sdk.PrefixValidator+sdk.PrefixConsensus, Bech32MainPrefix+sdk.PrefixValidator+sdk.PrefixConsensus+sdk.PrefixPublic)
+}
+
+func privKeyToBandAccAddress(privKey crypto.PrivKey) sdk.AccAddress {
+	config := sdk.GetConfig()
+	SetBech32AddressPrefixes(config)
+	return sdk.AccAddress(privKey.PubKey().Address().Bytes())
+}
+
+// NewBandProvider create new BandProvider create new cliCtx and txBldr
```
sorawit: creates

chain/provider/provider.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (provider *BandProvider) Sender() sdk.AccAddress {
+	return provider.addr
+}
+
+func (provider *BandProvider) SendTransaction(
+	msgs []sdk.Msg, seq, gas uint64,
+	memo, fees, gasPrices, broadcastMode string,
+) (sdk.TxResponse, error) {
+	cliCtx := provider.cliCtx.WithBroadcastMode(broadcastMode)
```
sorawit: What happens if someone passes a bogus `broadcastMode?`

chain/provider/utils.go
```diff
@@ -76,10 +76,5 @@ func completeAndBroadcastTxCLI(
 		return sdk.TxResponse{}, err
 	}
 	// broadcast to a Tendermint node
-	res, err := cliCtx.BroadcastTx(txBytes)
-	if err != nil {
-		return sdk.TxResponse{}, err
-	}
-
-	return res, nil
+	return cliCtx.BroadcastTx(txBytes)
```
sorawit: Good!

chain/provider/provider.go
```diff
@@ -0,0 +1,67 @@
+package provider
```
sorawit: Let's name this package `d3nlib` so we can have many things in this.

sorawit (changes requested):

chain/provider/provider.go
taobun: It will return error when broadcast transaction code in this [file](https://github.com/cosmos/cosmos-sdk/blob/master/client/context/broadcast.go#L31)

prin-r (approved):

sorawit (approved):
lgtm

PR #439, opened 2020-02-05
Implement `BandStatefulClient` package

taobun:
Now bandsv and provider use new client that can send request with different sequence number correctly (Can send 2 request in same block with correct sequence number i, i+1)

### DIFF
--- a/chain/d3nlib/client.go
+++ b/chain/d3nlib/client.go
@@ -1,0 +1,59 @@
+package d3nlib
+
+import (
+	"sync"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	"github.com/tendermint/tendermint/crypto"
+)
+
+type BandStatefulClient struct {
+	sequenceNumber uint64
+
+	mtx      sync.RWMutex
+	provider BandProvider
+}
+
+func NewBandStatefulClient(nodeURI string, privKey crypto.PrivKey) (BandStatefulClient, error) {
+	provider, err := NewBandProvider(nodeURI, privKey)
+	if err != nil {
+		return BandStatefulClient{}, err
+	}
+	_, seq, err := authtypes.NewAccountRetriever(provider.cliCtx).GetAccountNumberSequence(provider.Sender())
+	if err != nil {
+		return BandStatefulClient{}, err
+	}
+
+	return BandStatefulClient{
+		sequenceNumber: seq,
+		provider:       provider,
+	}, nil
+}
+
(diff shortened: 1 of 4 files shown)


chain/d3nlib/client.go
```diff
@@ -0,0 +1,67 @@
...
+	mtx      sync.RWMutex
+	provider BandProvider
+}
+
+func NewBandStatefulClient(privKey crypto.PrivKey) (BandStatefulClient, error) {
+	provider, err := NewBandProvider(privKey)
+	if err != nil {
+		return BandStatefulClient{}, err
+	}
+	_, seq, err := authtypes.NewAccountRetriever(provider.cliCtx).GetAccountNumberSequence(provider.Sender())
```
sorawit: Should this `GetAccountNumberSequence` be implemented in `BandProvider`?

sorawit (changes requested):

chain/d3nlib/client.go
```diff
@@ -0,0 +1,67 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	authtypes "github.com/cosmos/cosmos-sdk/x/auth/types"
+	"github.com/tendermint/tendermint/crypto"
+)
+
+type BandStatefulClient struct {
+	sequenceNumber uint64
+
+	mtx      sync.RWMutex
```
sorawit: Let's use `sync.Mutex` since we don't need the fancy read-write feature

chain/d3nlib/client.go
```diff
@@ -0,0 +1,67 @@
...
+	msg sdk.Msg, gas uint64,
+	memo, fees, gasPrices, broadcastMode string,
+) (sdk.TxResponse, error) {
+	// Ask current sequence number
+	_, seq, err := authtypes.NewAccountRetriever(client.provider.cliCtx).
+		GetAccountNumberSequence(client.provider.Sender())
+	if err != nil {
+		return sdk.TxResponse{}, err
+	}
+	client.mtx.Lock()
```
sorawit: This is fine as is since it's obvious that nothing between lock and unlock can fail. But you can also write it like this.

```
// YOUR OTHER CODE
{
     client.mtx.Lock()
     defer client.mtx.Unlock()
     // YOUR CODE HERE UNDER LOCK
}
// YOUR OTHER CODE
```

sorawit (changes requested):

taobun:
Review this pr too @sorawit @prin-r

sorawit (dismissed):

evilpeach:
Merge this before #442

evilpeach (approved):

prin-r (approved):

PR #452, opened 2020-02-07
fix 20 years in ui, use now unix instead of 0

prin-r:


sorawit:
I don't think this is how we should fix it. It will be confusing to user why the code creation time keeps changing. Let's fix it on the frontend.

PR #494, opened 2020-02-10
chain: Change default commit timeout to 1 second

### LINKED ISSUE #493: Change default propose timeout to 1 second
According to https://forum.cosmos.network/t/consensus-timeouts-explained/1421 and a talk with Cosmos dev, changing the timeout to 1 means the block producer will wait 1 sec *after* collecting > 2/3 of signatures to produce blocks. There shouldn't be any problem with that. We can revisit if testnet poses a problem.

sorawit:
Fix #493

### DIFF
--- a/chain/cmd/bandd/main.go
+++ b/chain/cmd/bandd/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"encoding/json"
 	"io"
+	"time"
 
 	"github.com/cosmos/cosmos-sdk/baseapp"
 	"github.com/cosmos/cosmos-sdk/client"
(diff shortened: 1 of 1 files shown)


chain/cmd/bandd/main.go
```diff
@@ -41,9 +42,16 @@ func main() {
...
-		Short:             "Band D3N App Daemon (server)",
-		PersistentPreRunE: server.PersistentPreRunEFn(ctx),
+		Use:   "bandd",
+		Short: "Band D3N App Daemon (server)",
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			err := server.PersistentPreRunEFn(ctx)(cmd, args)
+			if err != nil {
+				return err
+			}
+			ctx.Config.Consensus.TimeoutCommit = 1000 * time.Millisecond
```
prin-r: Why not time.Second

taobun (approved):

PR #520, opened 2020-02-12
chain/owasm: Add end-to-end testing code with silly wasm

sorawit:


chain/owasm/res/silly.rs
```diff
@@ -0,0 +1,44 @@
+extern "C" {
```
sorawit, line 1: This is for you to read the raw rust code. Not actually used.

chain/owasm/execute_test.go
```diff
@@ -13,7 +13,7 @@ func TestExecuteCanCallEnv(t *testing.T) {
 	if err != nil {
 		panic(err)
 	}
-	result, gasUsed, err := Execute(NewMockExecutionEnvironment(), code, "execute", []byte{}, 10000)
+	result, gasUsed, err := Execute(&mockExecutionEnvironment{}, code, "execute", []byte{}, 10000)
 	if err != nil {
```
taobun: Use `require.NotNil(t, err)` instead if and panic

chain/owasm/execute_test.go
```diff
@@ -26,6 +26,30 @@ func TestExecuteOutOfGas(t *testing.T) {
...
 		panic(err)
 	}
-	_, _, err = Execute(NewMockExecutionEnvironment(), code, "execute", []byte{}, 10)
+	_, _, err = Execute(&mockExecutionEnvironment{}, code, "execute", []byte{}, 10)
 	require.EqualError(t, err, "gas limit exceeded")
 }
+
+func TestExecuteEndToEnd(t *testing.T) {
+	code, err := ioutil.ReadFile("./res/silly.wasm")
+	if err != nil {
```
taobun: ditto

taobun (changes requested):

chain/owasm/execute_test.go
sorawit: Thanks! Should be `require.Nil` though.

chain/owasm/execute_test.go
```diff
@@ -10,10 +10,8 @@ import (
 
 func TestExecuteCanCallEnv(t *testing.T) {
 	code, err := ioutil.ReadFile("./res/main.wasm")
-	if err != nil {
-		panic(err)
-	}
-	result, gasUsed, err := Execute(NewMockExecutionEnvironment(), code, "execute", []byte{}, 10000)
+	require.Nil(t, err)
+	result, gasUsed, err := Execute(&mockExecutionEnvironment{}, code, "execute", []byte{}, 10000)
 	if err != nil {
```
taobun: This code too

taobun (changes requested):

chain/owasm/execute_test.go
sorawit: Right. Thanks!

taobun (approved):
Nice!

prin-r (approved):

PR #521, opened 2020-02-12
Request keeper

### LINKED ISSUE #495: Implement keeper and type for data requests and data reports
Design and implement keepers and data structures for data requests and reports. 

For data requests, it should store:
- The oracle script identifier
- The parameters of the request
- The list of validators that is responsible for responding to the request
- The list of validators that already submit report transaction
- The block height and block time that this request is prepared
- The block height that this request will expire
- Whether it has been resolved

For data reports, it should store (keyed by request id, external id, validator):
- The data from reporter

taobun:
fixed #495 

A lot of code in query part is commented. Because new request structure


chain/x/zoracle/internal/keeper/request.go
```diff
@@ -24,16 +24,53 @@ func (k Keeper) GetRequest(ctx sdk.Context, id uint64) (types.Request, sdk.Error
 	return request, nil
 }
 
+func (k Keeper) AddSubmitValidator(ctx sdk.Context, id int64, validator sdk.ValAddress) sdk.Error {
```
evilpeach: add comment

chain/x/zoracle/internal/keeper/request.go
```diff
@@ -24,16 +24,53 @@ func (k Keeper) GetRequest(ctx sdk.Context, id uint64) (types.Request, sdk.Error
...
+
+	if !found {
+		return types.ErrInvalidValidator(types.DefaultCodespace)
+	}
+	request.SubmittedValidatorList = append(request.SubmittedValidatorList, validator)
+	k.SetRequest(ctx, id, request)
+	return nil
+}
+
+func (k Keeper) SetResolve(ctx sdk.Context, id int64, isResolved bool) sdk.Error {
```
evilpeach, line 56: ditto

chain/x/zoracle/internal/types/request.go
```diff
@@ -1,32 +1,39 @@
...
-	Params      []byte `json:"params"`
-	ReportEndAt uint64 `json:"reportEnd"`
+	OracleScriptID           int64            `json:"oracleScriptID"`
+	Calldata                 []byte           `json:"calldata"`
+	ValidatorList            []sdk.ValAddress `json:"validatorList"`
+	SufficientValidatorCount int64            `json:"sufficientValidatorCount"`
+	SubmittedValidatorList   []sdk.ValAddress `json:"submittedValidatorList"`
+	RequestedBlock           int64            `json:"requestedBlock"`
+	RequestedTime            int64            `json:"requestedTime"`
+	ExpiredBlock             int64            `json:"expiredBlock"`
```
sorawit: Let's call there fields `ExpirationHeight` and `RequestHeight`(we use "request" here as an [Adjectival noun](https://en.wikipedia.org/wiki/Noun_adjunct) using "requested height" would mean the "height" is requested - that's not right).

chain/x/zoracle/internal/types/request.go
```diff
@@ -1,32 +1,39 @@
...
 
-// Request is a type to store detail of request
+// Request is a type to store detail of request.
 type Request struct {
-	CodeHash    []byte `json:"codeHash"`
-	Params      []byte `json:"params"`
-	ReportEndAt uint64 `json:"reportEnd"`
+	OracleScriptID           int64            `json:"oracleScriptID"`
+	Calldata                 []byte           `json:"calldata"`
+	ValidatorList            []sdk.ValAddress `json:"validatorList"`
```
sorawit: Two things.
1. For `ValidatorList`, let's just call it `Validators` or `ActiveValidators`. Using word `List` when the type already tells you it's a list is bit redundant.
2. Do we need `SubmittedValidatorList`? I think we can just iterate on all reports, right? Maybe we just need `ReportCount`?

chain/x/zoracle/internal/types/request.go
```diff
@@ -1,32 +1,39 @@
 package types
 
 import (
-	"fmt"
-	"strings"
+	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-// Request is a type to store detail of request
+// Request is a type to store detail of request.
```
sorawit: Request is a data structure that stores the detail of a request to an oracle script.

chain/x/zoracle/internal/keeper/request_test.go
```diff
@@ -4,49 +4,150 @@ import (
...
+func TestAddInvalidSubmitValidator(t *testing.T) {
+	ctx, keeper := CreateTestInput(t, false)
+	request := newDefaultRequest()
 
 	keeper.SetRequest(ctx, 1, request)
+
+	err := keeper.AddSubmitValidator(ctx, 1, sdk.ValAddress([]byte("validator3")))
+	require.Equal(t, types.CodeInvalidValidator, err.Code())
+
 	actualRequest, err := keeper.GetRequest(ctx, 1)
```
evilpeach, line 73: check `validator3` is not in `actualRequest` ?

chain/x/zoracle/internal/types/report.go
```diff
@@ -53,3 +31,23 @@ func NewExternalData(externalDataID int64, data []byte) ExternalData {
 		Data:           data,
 	}
 }
+
+// ValidatorReport is a report that contain operator address in struct
+type ValidatorReport struct {
```
sorawit: Let's call this `ReportWithValidator` for extra clarity.

chain/x/zoracle/internal/keeper/request.go
sorawit, line 56: Feels really bad to load the whole struct just to set one boolean value.... But I think this is the standard. So ok.

chain/x/zoracle/internal/keeper/report.go
```diff
@@ -3,27 +3,31 @@ package keeper
...
-func (k Keeper) GetReportsIterator(ctx sdk.Context, requestID uint64) sdk.Iterator {
+func (k Keeper) GetReportsIterator(ctx sdk.Context, requestID int64) sdk.Iterator {
 	prefix := types.GetIteratorPrefix(types.ReportKeyPrefix, requestID)
 	store := ctx.KVStore(k.storeKey)
 	return sdk.KVStorePrefixIterator(store, prefix)
 }
 
 // GetDataReports returns all the reports for a specific request ID.
-func (k Keeper) GetDataReports(ctx sdk.Context, requestID uint64) []types.Report {
+func (k Keeper) GetDataReports(ctx sdk.Context, requestID int64) []types.Report {
```
sorawit: Some of these might not be needed anymore. We will see after we finish the handler (:

sorawit (changes requested):

chain/x/zoracle/internal/keeper/request_test.go
taobun: Test equal in line 75 already

chain/x/zoracle/internal/types/request.go
```diff
@@ -1,32 +1,39 @@
...
-	ReportEndAt uint64 `json:"reportEnd"`
+	OracleScriptID           int64            `json:"oracleScriptID"`
+	Calldata                 []byte           `json:"calldata"`
+	ValidatorList            []sdk.ValAddress `json:"validatorList"`
+	SufficientValidatorCount int64            `json:"sufficientValidatorCount"`
+	SubmittedValidatorList   []sdk.ValAddress `json:"submittedValidatorList"`
+	RequestedBlock           int64            `json:"requestedBlock"`
+	RequestedTime            int64            `json:"requestedTime"`
+	ExpiredBlock             int64            `json:"expiredBlock"`
+	IsResolved               bool             `json:"isResolved"`
```
sorawit, line 18: As we already talked, I think we'll need to have a way to track the list of external data requests. Each will have the external ID, the data source ID, and calldata.

sorawit (changes requested):

chain/x/zoracle/internal/types/request.go
sorawit: Ok so we decide to keep both lists. Let's name it `RequestedValidators` and `ReceivedValidators`.

sorawit (changes requested):

chain/x/zoracle/internal/keeper/raw_request.go
```diff
@@ -0,0 +1,30 @@
+package keeper
+
+import (
+	"github.com/bandprotocol/d3n/chain/x/zoracle/internal/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// SetRawDataRequest is a function to save raw data request detail to the given request id and external id.
+func (k Keeper) SetRawDataRequest(ctx sdk.Context, requestID, externalID, scriptID int64, calldata []byte) {
```
sorawit: shouldn't this be data source id, not script id?

chain/x/zoracle/internal/types/key.go
```diff
@@ -28,48 +28,57 @@ var (
...
+
+	// RawDataReportStoreKeyPrefix is a prefix for report store
+	RawDataReportStoreKeyPrefix = []byte{0x03}
 )
 
 // RequestStoreKey is a function to generate key for each request in store
-func RequestStoreKey(requestID uint64) []byte {
-	buf := uint64ToBytes(requestID)
+func RequestStoreKey(requestID int64) []byte {
+	buf := int64ToBytes(requestID)
```
sorawit: Make it fabulous

chain/x/zoracle/internal/types/key.go
```diff
@@ -28,48 +28,57 @@ var (
...
 	return append(RequestStoreKeyPrefix, buf...)
 }
 
 // ResultStoreKey is a function to generate key for each result in store
-func ResultStoreKey(requestID uint64, codeHash []byte, params []byte) []byte {
-	buf := uint64ToBytes(requestID)
+func ResultStoreKey(requestID int64, codeHash []byte, params []byte) []byte {
```
sorawit, line 50: Let's make it more readable and follow the same pattern.

chain/x/zoracle/internal/types/key.go
```diff
@@ -28,48 +28,57 @@ var (
...
+	buf := append(RawDataReportStoreKeyPrefix, int64ToBytes(requestID)...)
+	buf = append(buf, int64ToBytes(externalID)...)
 	return append(buf, validatorAddress.Bytes()...)
 }
 
 // GetIteratorPrefix is a function to get specific prefix
-func GetIteratorPrefix(prefix []byte, requestID uint64) []byte {
-	buf := uint64ToBytes(requestID)
+func GetIteratorPrefix(prefix []byte, requestID int64) []byte {
+	buf := int64ToBytes(requestID)
```
sorawit: Make it fabulous.

sorawit (changes requested):

chain/x/zoracle/internal/types/request.go
```diff
@@ -1,32 +1,54 @@
...
-	return strings.TrimSpace(fmt.Sprintf(`CodeHash: %x
-Params: %x
-ReportEndAt: %d`,
-		req.CodeHash,
-		req.Params,
-		req.ReportEndAt,
-	))
+// RawDataRequest is a data structure that store what datasource and calldata will be used in request.
+type RawDataRequest struct {
+	ScriptID int64  `json:"scriptID"`
```
sorawit: DataSourceID. Not ScriptID.

sorawit (approved):

evilpeach (approved):

PR #522, opened 2020-02-12
remove bsv key from scan because binance is not support bsv, so we remove it and add etc

### LINKED ISSUE #514: Update BSV enum on scan to ETC

prin-r:
Fix #514


sorawit (approved):

evilpeach (approved):
LGTM

PR #530, opened 2020-02-13
chain: keeper type oracle script

### LINKED ISSUE #492: Implement keeper and type for oracle scripts
This will be quite similar to the existing code, but will be referenced with an incremental ID as opposed to hash value. The work should also be quite similar to #491. An oracle script contains `Code`, `Name`, and `Owner`.

evilpeach:
Fixes #492

### DIFF
--- a/chain/x/zoracle/internal/keeper/keeper.go
+++ b/chain/x/zoracle/internal/keeper/keeper.go
@@ -76,3 +76,29 @@ func (k Keeper) GetNextDataSourceID(ctx sdk.Context) int64 {
 	store.Set(types.DataSourceCountStoreKey, bz)
 	return dataSourceCount + 1
 }
+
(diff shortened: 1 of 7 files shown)


taobun (approved):

chain/x/zoracle/internal/keeper/keeper.go
```diff
@@ -76,3 +76,29 @@ func (k Keeper) GetNextDataSourceID(ctx sdk.Context) int64 {
 	store.Set(types.DataSourceCountStoreKey, bz)
 	return dataSourceCount + 1
 }
+
+// GetOracleScriptCount returns the current number of all oracle script ever exist.
```
sorawit, line 80: all oracle scripts. plural.

sorawit (approved):

PR #531, opened 2020-02-13
Introduce module level parameters for zoracle

prin-r:
Fix #490

PR #532, opened 2020-02-13
Implement execution environment

### LINKED ISSUE #500: Implement execution environment, both during prepare and aggregate phase
This will involve both designing the data structure and make sure it implements proper interface

taobun:
fixed #500

### DIFF
--- a/chain/x/zoracle/exec_env.go
+++ b/chain/x/zoracle/exec_env.go
@@ -54,7 +54,7 @@ func (env *ExecutionEnvironment) GetAggregateBlockTime() int64 {
 }
 
 func (env *ExecutionEnvironment) GetValidatorAddress(validatorIndex int64) ([]byte, error) {
-	if validatorIndex >= int64(len(env.request.RequestedValidators)) {
+	if validatorIndex < 0 || validatorIndex >= int64(len(env.request.RequestedValidators)) {
 		return []byte{}, errors.New("validator out of range")
 	}
 	return env.request.RequestedValidators[validatorIndex], nil
@@ -73,7 +73,7 @@ func (env *ExecutionEnvironment) GetExternalData(
 	externalDataID int64,
 	validatorIndex int64,
 ) ([]byte, error) {
-	if validatorIndex >= int64(len(env.request.RequestedValidators)) {
+	if validatorIndex < 0 || validatorIndex >= int64(len(env.request.RequestedValidators)) {
 		return []byte{}, errors.New("validator out of range")
 	}
 	validatorAddress := env.request.RequestedValidators[validatorIndex]
--- a/chain/x/zoracle/exec_env.go
+++ b/chain/x/zoracle/exec_env.go
@@ -77,9 +77,5 @@ func (env *ExecutionEnvironment) GetExternalData(
 	}
 	validatorAddress := env.request.RequestedValidators[validatorIndex]
 
-	report, err := env.keeper.GetRawDataReport(env.ctx, env.requestID, externalDataID, validatorAddress)
-	if err != nil {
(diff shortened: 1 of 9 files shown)


chain/x/zoracle/exec_env.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (env *ExecutionEnvironment) GetAggregateBlockTime() int64 {
+	if int64(len(env.request.ReceivedValidators)) >= env.request.SufficientValidatorCount {
+		return env.ctx.BlockTime().Unix()
+	}
+	return 0
+}
+
+func (env *ExecutionEnvironment) GetValidatorAddress(validatorIndex int64) ([]byte, error) {
+	if validatorIndex >= int64(len(env.request.RequestedValidators)) {
```
sorawit: What happens if we get a negative number?

sorawit: Same question for the two functions below.

chain/x/zoracle/exec_env.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	return env.request.RequestedValidators[validatorIndex], nil
+}
+
+func (env *ExecutionEnvironment) RequestExternalData(
+	dataSourceID int64,
+	externalDataID int64,
+	calldata []byte,
+) error {
+	env.keeper.SetRawDataRequest(env.ctx, env.requestID, externalDataID, dataSourceID, calldata)
```
sorawit: This must fail if the dataSourceID does not exist.

sorawit (changes requested):

chain/x/zoracle/exec_env.go
sorawit: Must also fail if externalData is already used

sorawit (changes requested):

chain/x/zoracle/exec_env.go
sorawit: Must also fail if the calldata size is too big. Pending on #531

sorawit (changes requested):

chain/x/zoracle/exec_env.go
```diff
@@ -0,0 +1,85 @@
...
+func (env *ExecutionEnvironment) GetExternalData(
+	externalDataID int64,
+	validatorIndex int64,
+) ([]byte, error) {
+	if validatorIndex < 0 || validatorIndex >= int64(len(env.request.RequestedValidators)) {
+		return []byte{}, errors.New("validator out of range")
+	}
+	validatorAddress := env.request.RequestedValidators[validatorIndex]
+
+	report, err := env.keeper.GetRawDataReport(env.ctx, env.requestID, externalDataID, validatorAddress)
```
sorawit: Why don't we just `return` here?

chain/x/zoracle/internal/keeper/raw_request.go
```diff
@@ -28,3 +47,8 @@ func (k Keeper) GetRawDataRequest(
 	k.cdc.MustUnmarshalBinaryBare(bz, &requestDetail)
 	return requestDetail, nil
 }
+
+func (k Keeper) CheckRawDataRequestExists(ctx sdk.Context, requestID, externalID int64) bool {
```
sorawit, line 53: Comment?

sorawit (changes requested):

chain/x/zoracle/exec_env.go
taobun: Good catch!

sorawit (approved):
Looks good

evilpeach (approved):

PR #544, opened 2020-02-14
chain: Request handler

taobun:


chain/x/zoracle/internal/keeper/request.go
```diff
@@ -24,6 +24,51 @@ func (k Keeper) GetRequest(ctx sdk.Context, id int64) (types.Request, sdk.Error)
 	return request, nil
 }
 
+// Request attempts to create new request.
+// An error is returned if some conditions failed
+func (k Keeper) Request(
```
sorawit: Let's call this `AddRequest` to be on par with the ones in data source keeper and oracle keeper.

chain/x/zoracle/internal/keeper/raw_request.go
```diff
@@ -54,3 +38,57 @@ func (k Keeper) CheckRawDataRequestExists(ctx sdk.Context, requestID, externalID
...
+	// Check data source exist
+	if !k.CheckDataSourceExists(ctx, dataSourceID) {
+		// TODO: fix error later
+		return types.ErrRequestNotFound(types.DefaultCodespace)
+	}
+	if k.CheckRawDataRequestExists(ctx, requestID, externalID) {
+		// TODO: fix error later
+		return types.ErrRequestNotFound(types.DefaultCodespace)
+	}
+	k.SetRawDataRequest(ctx, requestID, externalID, types.NewRawDataRequest(dataSourceID, calldata))
```
sorawit, line 62: This should throw an error back if the number of raw data request exceeds the maximum number. I know we already check in the handler but returning an error here allows the oracle script to act upon the error.

chain/x/zoracle/internal/keeper/raw_request.go
```diff
@@ -54,3 +38,57 @@ func (k Keeper) CheckRawDataRequestExists(ctx sdk.Context, requestID, externalID
 	store := ctx.KVStore(k.storeKey)
 	return store.Has(types.RawDataRequestStoreKey(requestID, externalID))
 }
+
+// AddNewRawDataRequest check all conditions before save new raw data request to store.
```
sorawit: // AddNewRawDataRequest checks all conditions before saving a new raw data request to the store.

chain/x/zoracle/internal/keeper/raw_request.go
```diff
@@ -54,3 +38,57 @@ func (k Keeper) CheckRawDataRequestExists(ctx sdk.Context, requestID, externalID
...
+func (k Keeper) AddNewRawDataRequest(
+	ctx sdk.Context, requestID, externalID, dataSourceID int64, calldata []byte,
+) sdk.Error {
+	// TODO: Check calldata size
+
+	// Check request exist
```
sorawit: Choose one like a druid without Fandral Staghelm: 
1. remove the comment (since the code is quite clear already)
2. update comment to be more of a full sentence: // Check that the request exists in the store. 

Repeat for everything below.

chain/x/zoracle/internal/keeper/request.go
```diff
@@ -24,6 +24,51 @@ func (k Keeper) GetRequest(ctx sdk.Context, id int64) (types.Request, sdk.Error)
...
+		sufficientValidatorCount,
+		ctx.BlockHeight(),
+		ctx.BlockTime().Unix(),
+		ctx.BlockHeight()+expiration,
+	))
+
+	return requestID, nil
+}
+
+// ValidateDataSourceCount validate amount of raw data requests
```
sorawit: validates that the number of ... is not

chain/x/zoracle/internal/keeper/request.go
```diff
@@ -70,6 +115,15 @@ func (k Keeper) CheckRequestExists(ctx sdk.Context, id int64) bool {
 	return store.Has(types.RequestStoreKey(id))
 }
 
+// HasToPutInPendingList return boolean that request must be put in pending list or not.
```
sorawit: // returns whether the request can be moved to the pending resolve list.

Let's name this function `ShouldResolve`

chain/x/zoracle/internal/keeper/request.go
```diff
@@ -24,6 +24,51 @@ func (k Keeper) GetRequest(ctx sdk.Context, id int64) (types.Request, sdk.Error)
 	return request, nil
 }
 
+// Request attempts to create new request.
+// An error is returned if some conditions failed
```
sorawit: // Request attempts to create a new request. An error is returned if some conditions failed.

sorawit (changes requested):

chain/x/zoracle/internal/keeper/request.go
sorawit: // ShouldBecomePendingResolve checks and returns whether the given request should be moved to the 
// pending resolve list, which will be resolved during the EndBlock call. The move will happen exactly when
// the request receives sufficient raw reports from the validators.

Ok. Let's call this `ShouldBecomePendingResolve`. We will call the list `PendingResolveList`

sorawit (changes requested):

chain/x/zoracle/internal/keeper/request.go
```diff
@@ -24,6 +24,50 @@ func (k Keeper) GetRequest(ctx sdk.Context, id int64) (types.Request, sdk.Error)
...
+		sufficientValidatorCount,
+		ctx.BlockHeight(),
+		ctx.BlockTime().Unix(),
+		ctx.BlockHeight()+expiration,
+	))
+
+	return requestID, nil
+}
+
+// ValidateDataSourceCount validate that the number of raw data requests is
```
sorawit: validates

chain/x/zoracle/internal/keeper/request.go
```diff
@@ -93,8 +148,8 @@ func (k Keeper) SetPendingRequests(ctx sdk.Context, reqIDs []int64) {
 	store.Set(types.UnresolvedRequestListStoreKey, encoded)
 }
 
-// GetPendingRequests returns the list of pending request.
-func (k Keeper) GetPendingRequests(ctx sdk.Context) []int64 {
+// GetPendingResolveList returns the list of pending request.
+func (k Keeper) GetPendingResolveList(ctx sdk.Context) []int64 {
 	store := ctx.KVStore(k.storeKey)
 	reqIDsBytes := store.Get(types.UnresolvedRequestListStoreKey)
```
sorawit: While you are here. Let's update the key name too.

sorawit (changes requested):

sorawit (approved):

evilpeach (approved):

PR #549, opened 2020-02-14
Allow owasm to run test without needing to commit wasm code

### LINKED ISSUE #475: Allow owasm to run test without needing to commit .wasm code
Currently to test owasm code, ones would need to commit `.wasm.` which is huge and not simple to see changes. With this ticket, we expect to have test resources written in either high level languages (c++, rust, go, etc) or wat. The test and CI should still be working.

prin-r:
Fix #475


chain/test.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+source ~/.profile
```
sorawit: Remove this line. Whoever running this should know to `source` themselves.

chain/.gitlab-ci.yml
```diff
@@ -1,9 +1,9 @@
 chain-test:
   stage: test
-  image: golang:1.13
+  image: bandprotocol/go-rust:0.1
   script:
     - cd chain
-    - go test ./...
+    - chmod +x ./test.sh && ./test.sh
```
sorawit: `chmod` should be done before committing the code.

chain/test.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
sorawit: Call this `run_tests`

sorawit (changes requested):

chain/.gitignore
```diff
@@ -9,3 +9,8 @@
 vendor
 .vendor-new
 build
+
+Cargo.lock
```
sorawit: should Cargo.lock be ignored? @taobun

prin-r: The purpose of a Cargo.lock is to describe the state of the world at the time of a successful build. It is then used to provide deterministic builds across whatever machine is building the package by ensuring that the exact same dependencies are being compiled.

This property is most desirable from applications and packages which are at the very end of the dependency chain (binaries). As a result, it is recommended that all binaries check in their Cargo.lock.

For libraries the situation is somewhat different. A library is not only used by the library developers, but also any downstream consumers of the library. Users dependent on the library will not inspect the library’s Cargo.lock (even if it exists). This is precisely because a library should not be deterministically recompiled for all users of the library.

If a library ends up being used transitively by several dependencies, it’s likely that just a single copy of the library is desired (based on semver compatibility). If Cargo used all of the dependencies' Cargo.lock files, then multiple copies of the library could be used, and perhaps even a version conflict.

In other words, libraries specify semver requirements for their dependencies but cannot see the full picture. Only end products like binaries have a full picture to decide what versions of dependencies should be used.

https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries

prin-r: From my understanding we should add Cargo.lock

chain/owasm/res/main/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+extern "C" {
+    pub fn getCurrentRequestID() -> i64;
+    pub fn getRequestedValidatorCount() -> i64;
+
+    pub fn saveReturnData(dataOffset: *const u8, dataLength: i32);
+}
+
+#[no_mangle]
+pub extern "C" fn execute() {
```
taobun: We already changed to wasm-pack. `pub extern "C"` is needed to use?

prin-r: done

taobun (approved):

sorawit (approved):

PR #591, opened 2020-02-17
Rest-server request and report index endpoint

### LINKED ISSUE #557: Rest-server request and report index endpoint
- [x] Querier types/keeper to get a request and reports from the store.
- [x] A REST endpoint to get the request and reports given the ID.

taobun:
fixed #557


chain/x/zoracle/client/rest/query.go
```diff
@@ -13,93 +13,98 @@ import (
...
+
+	request.OracleScriptID = queryRequest.Request.OracleScriptID
+	request.Calldata = queryRequest.Request.Calldata
+	request.RequestedValidators = queryRequest.Request.RequestedValidators
+	request.SufficientValidatorCount = queryRequest.Request.SufficientValidatorCount
+	request.ExpirationHeight = queryRequest.Request.ExpirationHeight
+	request.IsResolved = queryRequest.Request.IsResolved
+
+	request.Result = queryRequest.Result
+	if queryRequest.Result == nil {
```
sorawit: I think we can leave it as `nil`. People that call this function should know that if `isResolved` is false then result is `nil`.

chain/x/zoracle/client/rest/types.go
```diff
@@ -17,22 +14,24 @@ type ScriptInfoWithTx struct {
...
-	CodeHash          cmn.HexBytes     `json:"codeHash"`
-	Params            json.RawMessage  `json:"params"`
-	TargetBlock       int64            `json:"targetBlock"`
-	Requester         sdk.AccAddress   `json:"requester"`
-	TxHash            string           `json:"txhash"`
-	RequestedAtHeight int64            `json:"requestedAtHeight"`
-	RequestedAtTime   string           `json:"requestedAtTime"`
-	Reports           []ReportDetail   `json:"reports"`
-	Result            json.RawMessage  `json:"result"`
+	OracleScriptID           int64            `json:"oracleScriptID"`
```
sorawit: I think we should also return raw data requests.

chain/x/zoracle/internal/types/querier.go
```diff
@@ -23,35 +23,26 @@ func (u64a U64Array) String() string {
...
-// type RequestInfo struct {
-// 	CodeHash    []byte            `json:"codeHash"`
-// 	Params      json.RawMessage   `json:"params"`
-// 	ParamsRaw   []byte            `json:"paramsRaw"`
-// 	TargetBlock uint64            `json:"targetBlock"`
-// 	Reports     []ReportWithValidator `json:"reports"`
-// 	Result      json.RawMessage   `json:"result"`
-// 	ResultRaw   []byte            `json:"resultRaw"`
-// }
+type RequestInfo struct {
```
sorawit: Let's call this `RequestQuerierInfo`

sorawit (changes requested):

chain/x/zoracle/client/rest/types.go
sorawit: Let's call this `RequestRESTInfo`

chain/x/zoracle/internal/types/querier.go
```diff
@@ -23,35 +23,26 @@ func (u64a U64Array) String() string {
...
-// 		CodeHash:    codeHash,
-// 		ParamsRaw:   paramsRaw,
-// 		Params:      params,
-// 		TargetBlock: targetBlock,
-// 		Reports:     reports,
-// 		Result:      result,
-// 		ResultRaw:   resultRaw,
-// 	}
-// }
+func NewRequestInfo(
```
sorawit: Also update the name here.

chain/x/zoracle/client/rest/query.go
```diff
@@ -13,93 +13,98 @@ import (
 	"github.com/bandprotocol/d3n/chain/x/zoracle/internal/types"
 )
 
+func newRequestQueryInfo(
```
sorawit:
```
// buildRequestRESTInfo takes a RequestQuerierInfo instance and builds a more comprehensive version of it.
func buildRequestRESTInfo
```

sorawit (changes requested):

chain/x/zoracle/client/rest/query.go
```diff
@@ -13,93 +13,96 @@ import (
...
+	request.RequestedValidators = queryRequest.Request.RequestedValidators
+	request.SufficientValidatorCount = queryRequest.Request.SufficientValidatorCount
+	request.ExpirationHeight = queryRequest.Request.ExpirationHeight
+	request.IsResolved = queryRequest.Request.IsResolved
+	request.RawDataRequests = queryRequest.RawDataRequests
+
+	request.Result = queryRequest.Result
+
+	// Get request detail
+	searchRequest, err := utils.QueryTxsByEvents(
```
sorawit, line 43: Can we somehow abstract "finding the transaction of this" as a separate function. We will need to reuse this a lot.

chain/x/zoracle/client/rest/types.go
```diff
@@ -17,22 +14,25 @@ type ScriptInfoWithTx struct {
...
+	OracleScriptID           int64                  `json:"oracleScriptID"`
+	Calldata                 []byte                 `json:"calldata"`
+	RequestedValidators      []sdk.ValAddress       `json:"requestedValidators"`
+	SufficientValidatorCount int64                  `json:"sufficientValidatorCount"`
+	ExpirationHeight         int64                  `json:"expirationHeight"`
+	IsResolved               bool                   `json:"isResolved"`
+	Requester                sdk.AccAddress         `json:"requester"`
+	TxHash                   string                 `json:"txhash"`
+	RequestedAtHeight        int64                  `json:"requestedAtHeight"`
+	RequestedAtTime          string                 `json:"requestedAtTime"`
```
sorawit: This can be put into its own struct if it helps with abstracting the transaction. The thing I mentioned above.

chain/x/zoracle/client/rest/types.go
```diff
@@ -17,22 +14,25 @@ type ScriptInfoWithTx struct {
...
-	ReportedAtTime   string          `json:"reportedAtTime"`
-	Value            json.RawMessage `json:"value"`
+	Reporter         sdk.ValAddress        `json:"reporter"`
+	TxHash           string                `json:"txhash"`
+	ReportedAtHeight int64                 `json:"reportedAtHeight"`
+	ReportedAtTime   string                `json:"reportedAtTime"`
+	Value            []types.RawDataReport `json:"value"`
 }
 
 type RequestQueryInfo struct {
```
sorawit: Let's call this RequestRESTInfo so we know it's the info for REST. We have way too many "info" structs lol.

sorawit (changes requested):

chain/x/zoracle/client/rest/query.go
```diff
@@ -6,100 +6,118 @@ import (
...
+		1,
+		10000, // Estimated validator reports
+	)
+
+	request.Reports = make([]ReportDetail, 0)
+
+	for _, report := range searchReports.Txs {
+		var validatorAddress sdk.ValAddress
+		for _, msg := range report.Tx.GetMsgs() {
+			if msg.Type() == "report" && msg.Route() == types.RouterKey {
```
sorawit: Should we use `MsgReportData{}.Type()` here? 

or alternatively should we just try to cast and check if it's ok? like
```
msgReport, ok := msg.(types.MsgReportData)
if ok {
    ...
}
```

Basically so we don't need to hardcode "report"

chain/x/zoracle/internal/keeper/querier.go
```diff
@@ -31,59 +34,58 @@ func NewQuerier(keeper Keeper) sdk.Querier {
...
-// }
+	reports := make([]types.ReportWithValidator, 0)
+	for validator, report := range reportMap {
+		reports = append(reports, types.NewReportWithValidator(
+			report, []byte(validator),
+		))
+	}
+
+	result, sdkErr := keeper.GetResult(ctx, id, request.OracleScriptID, request.Calldata)
+	if sdkErr != nil {
```
sorawit: Here we should leave it as nil since the `isResolved` is false?

sorawit (changes requested):
looks good overall. just a few comments.

sorawit (dismissed):

prin-r (dismissed):

sorawit (approved):

prin-r (approved):

PR #593, opened 2020-02-18
Remove DataSourceCount in Request struct

taobun:
Remove don't use the field if you want to know data source count, you can call `GetRawDataRequestCount` at `keeper/raw_request.go`.


sorawit (approved):
Looks good.

prin-r (approved):

PR #635, opened 2020-02-25
chain: Add max result param

### LINKED ISSUE #597: Add MaxResultSize parameter

evilpeach:
Fixes #597 
- [x] Add MaxResultSize in params and alias
- [x] Set & Get MaxResultSize in keeper and test case
- [x] AddResult function in result keeper and test case 
- [x] Use AddResult in handler and update test case


chain/x/zoracle/internal/types/params.go
```diff
@@ -29,6 +29,10 @@ const (
 	// The maximum size of raw data report per data source.
 	// Default value is set to 1 kB.
 	DefaultMaxRawDataReportSize = int64(1 * 1024)
+
+	// The maximum size of result from validator.
```
sorawit: not from validator....

chain/x/zoracle/internal/types/params.go
```diff
@@ -73,8 +81,9 @@ func (p Params) String() string {
   MaxOracleScriptCodeSize:      %d
   MaxCalldataSize:              %d
   MaxDataSourceCountPerRequest: %d
-  MaxRawDataReportSize:         %d
-`, p.MaxDataSourceExecutableSize, p.MaxOracleScriptCodeSize, p.MaxCalldataSize, p.MaxDataSourceCountPerRequest, p.MaxRawDataReportSize)
+	MaxRawDataReportSize:					%d
```
sorawit: Why not keep the same indentation style?

sorawit (changes requested):

chain/x/zoracle/internal/types/params.go
taobun: after execution

sorawit (approved):

taobun (approved):

PR #650, opened 2020-02-27
chain-rest: upgrade get latest txs

### LINKED ISSUE #596: Implement better REST latest txs handler
The code should try to binary search first to try to find the block. Talk to @sorawit for details.

evilpeach:
Fixes #596 

[Example1](http://rpc.alpha-debug.bandchain.org:1317/d3n/txs/latest?page=1&limit=100)
[Example2](http://scan-alpha.surge.sh/)

### DIFF
--- a/chain/client/client.go
+++ b/chain/client/client.go
@@ -87,7 +87,7 @@ func LatestTxsRequestHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
 
 		// limit maximum must be 100.
 		if limit > 100 {
-			rest.WriteErrorResponse(w, http.StatusBadRequest, "limit must be less than or equal 100.")
+			rest.WriteErrorResponse(w, http.StatusBadRequest, "limit must not be greater than 100.")
 			return
 		}
 
--- a/chain/client/client.go
+++ b/chain/client/client.go
@@ -87,7 +87,8 @@ func LatestTxsRequestHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
 
 		// limit maximum must be 100.
 		if limit > 100 {
-			limit = 100
+			rest.WriteErrorResponse(w, http.StatusBadRequest, "limit must be less than or equal 100.")
+			return
 		}
 
 		// TODO: (1) Sort result in desc order after tendermint/tendermint:#4253 is released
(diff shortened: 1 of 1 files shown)


sorawit:
This is not binary search implementation, but let's start with this first and see how it goes. At least it should now not break. It might get slow if we have LOTS of txs and we will figure that out later. Binary search can be used to find the highest height with enough txs and reuse pretty much the same code for finding txs.

chain/client/client.go
```diff
@@ -90,26 +90,52 @@ func LatestTxsRequestHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
...
-		for ; startIdx >= endIdx && startIdx >= 0; startIdx-- {
-			result.Txs = append(result.Txs, searchResult.Txs[startIdx])
+
+		if startIndex < 1 {
+			startIndex = 1
 		}
+
+		if endIndex < 1 {
+			result.Count = 0
+			rest.PostProcessResponseBare(w, cliCtx, result)
```
sorawit, line 117: I think page total should also be in here. Maybe set it above?

sorawit (changes requested):

chain/client/client.go
```diff
@@ -90,26 +90,52 @@ func LatestTxsRequestHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
...
+			return
+		}
+
+		todoIndex := startIndex
+		for todoIndex <= endIndex {
+			pageOfTodoIndex := (todoIndex-1)/limit + 1
```
taobun, line 123: If `limit` is more than 100, it leads to infinite loop.

taobun (changes requested):

chain/client/client.go
```diff
@@ -85,31 +85,62 @@ func LatestTxsRequestHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
 			return
 		}
 
+		// limit maximum must be 100.
+		if limit > 100 {
```
taobun, line 88: Return error

taobun (changes requested):

chain/client/client.go
```diff
@@ -85,31 +85,63 @@ func LatestTxsRequestHandlerFn(cliCtx context.CLIContext) http.HandlerFunc {
...
+			rest.WriteErrorResponse(w, http.StatusBadRequest, "limit must be less than or equal 100.")
```
sorawit: must not be greater than

sorawit (changes requested):

taobun (approved):

sorawit (approved):

PR #690, opened 2020-02-28
Update new Wasm gas policy

### LINKED ISSUE #582: Update owasm gas meter to use more than just the default of 1 for all operations
Can use [this](https://ewasm.readthedocs.io/en/mkdocs/determining_wasm_gas_costs/) as a reference.

taobun:
fixed #582


chain/owasm/gas_policy.go
```diff
@@ -0,0 +1,224 @@
...
+func NewDefaultWasmGasPolicy() WasmGasPolicy {
+	return WasmGasPolicy{
+		GasMap: defaultGas,
+	}
+}
+
+func (p *WasmGasPolicy) GetCost(key compiler.Instr) int64 {
+	gasCost, found := p.GasMap[key.Op]
+	if !found {
+		panic(fmt.Sprintf("Cannot found instruction %s", key.Op))
```
sorawit: If we panic here. I assume running wasm's Execution function will return error, correct?

chain/owasm/gas_policy.go
```diff
@@ -0,0 +1,224 @@
...
+	"f32.convert_s/i64":   3,
+	"f64.convert_s/i32":   3,
+	"f64.convert_s/i64":   3,
+
+	// Type-parametric operators.
+	"drop":   3,
+	"select": 3,
+}
+
+type WasmGasPolicy struct {
```
sorawit: Let's call this `BandChainGasPolicy`

chain/owasm/gas_policy.go
```diff
@@ -0,0 +1,224 @@
...
+	GasMap map[string]int64
```
sorawit: Don't think we need this data structure at all. `GetCost` can just read from the map directly?

sorawit (changes requested):

chain/owasm/gas_policy.go
taobun: Nope, Good catch! I change to return UnknownGasCost instead.

sorawit (dismissed):

evilpeach (approved):

sorawit (approved):

PR #692, opened 2020-02-28
Add an endpoint to get the list of validators with evm compatible addresses

### LINKED ISSUE #586: Add an endpoint to get the list of validators with EVM compatible addresses
The endpoint will work very similarly to `/staking/validators` with the added field that shows the addresses for EVM bridges.

prin-r:
Fix #586

### DIFF
--- a/chain/client/client.go
+++ b/chain/client/client.go
@@ -230,7 +230,7 @@ func ServeSwaggerUI() http.Handler {
 func RegisterRoutes(cliCtx context.CLIContext, r *mux.Router) {
 	r.HandleFunc("/d3n/blocks/latest", LatestBlocksRequestHandlerFn(cliCtx)).Methods("GET")
 	r.HandleFunc("/d3n/txs/latest", LatestTxsRequestHandlerFn(cliCtx)).Methods("GET")
-	r.HandleFunc("/bandchain/evm-validators", GetValidators(cliCtx)).Methods("GET")
+	r.HandleFunc("/bandchain/evm-validators", GetEVMValidators(cliCtx)).Methods("GET")
 	r.HandleFunc(fmt.Sprintf("/d3n/proof/{%s}", requestIDTag), GetProofHandlerFn(cliCtx)).Methods("GET")
 	r.HandleFunc("/d3n/health_check", GetHealthStatus(cliCtx)).Methods("GET")
 	r.HandleFunc("/d3n/provider_status", GetProviderStatus(cliCtx)).Methods("GET")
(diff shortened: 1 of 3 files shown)


chain/client/client.go
```diff
@@ -230,6 +230,7 @@ func ServeSwaggerUI() http.Handler {
 func RegisterRoutes(cliCtx context.CLIContext, r *mux.Router) {
 	r.HandleFunc("/d3n/blocks/latest", LatestBlocksRequestHandlerFn(cliCtx)).Methods("GET")
 	r.HandleFunc("/d3n/txs/latest", LatestTxsRequestHandlerFn(cliCtx)).Methods("GET")
+	r.HandleFunc("/staking/validators", GetValidators(cliCtx)).Methods("GET")
```
sorawit: Won't this collide with the existing `/staking/validators`? Let's call this `/bandchain/evm-validators`

chain/client/validators.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/cosmos/cosmos-sdk/client/context"
+	"github.com/cosmos/cosmos-sdk/types/rest"
+)
+
+type ValidatorMinimal struct {
+	Address     string `json:"address"`
+	VotingPower int64  `json:"voting_power"`
+}
+
+type ValidatorsOnEthereum struct {
```
sorawit: Don't think this has anything specific with Ethereum.

chain/client/validators.go
```diff
@@ -0,0 +1,65 @@
...
+	BlockHeight int64              `json:"block_height"`
+	Validators  []ValidatorMinimal `json:"validators"`
+}
+
+func GetValidators(cliCtx context.CLIContext) http.HandlerFunc {
```
sorawit: Let's call this `GetEVMValidators`

sorawit (changes requested):

chain/client/validators.go
```diff
@@ -0,0 +1,65 @@
...
+			}
+
+			if pubkey, err := crypto.DecompressPubkey(pubKeyBytes[:]); err != nil {
+				rest.WriteErrorResponse(w, http.StatusInternalServerError, err.Error())
+				return
+			} else {
+				validatorsOnETH.Validators = append(
+					validatorsOnETH.Validators,
+					ValidatorMinimal{
+						Address:     fmt.Sprintf("0x%x", crypto.PubkeyToAddress(*pubkey)),
```
taobun: Can we use `crypto.PubkeyToAddress(*pubkey).String()` instead?

taobun (changes requested):

chain/client/validators.go
prin-r: done

chain/client/validators.go
```diff
@@ -0,0 +1,64 @@
...
+			return
+		}
+
+		validators, err := node.Validators(nil)
+		if err != nil {
+			rest.WriteErrorResponse(w, http.StatusInternalServerError, err.Error())
+			return
+		}
+
+		validatorsOnETH := ValidatorsMinimal{}
```
taobun: EVM or ETH just choose one

taobun (changes requested):

chain/client/validators.go
prin-r: done

sorawit (approved):

taobun (approved):

PR #705, opened 2020-03-03
Add caching layer for owasm resolver

Benzbeeb:


chain/owasm/resolver.go
```diff
@@ -123,25 +130,40 @@ func (r *resolver) resolveRequestExternalData(vm *exec.VirtualMachine) int64 {
 	return 0
 }
 
+func (r *resolver) getExternalDataFromCache(externalDataID int64, validatorIndex int64) ([]byte, error) {
```
sorawit, line 135: This won't work if it's called two times and error. The first call will return error but the second call won't.

chain/owasm/resolver.go
```diff
@@ -123,25 +130,40 @@ func (r *resolver) resolveRequestExternalData(vm *exec.VirtualMachine) int64 {
...
 	if err != nil {
 		return -1
 	}
 	return int64(len(externalData))
 }
 
 func (r *resolver) resolveReadExternalData(vm *exec.VirtualMachine) int64 {
-	dataSourceID := GetLocalInt64(vm, 0)
-	externalDataID := GetLocalInt64(vm, 1)
+	externalDataID := GetLocalInt64(vm, 0)
```
sorawit, line 164: Was this a bug before? Nice catch!

chain/owasm/resolver.go
```diff
@@ -123,25 +130,40 @@ func (r *resolver) resolveRequestExternalData(vm *exec.VirtualMachine) int64 {
...
+	return externalData, err
+}
+
 func (r *resolver) resolveGetExternalDataSize(vm *exec.VirtualMachine) int64 {
 	externalDataID := GetLocalInt64(vm, 0)
 	validatorIndex := GetLocalInt64(vm, 1)
-	// TODO: ExternalData should be cached for both this function and the one below.
-	externalData, err := r.env.GetExternalData(externalDataID, validatorIndex)
+
+	externalData, err := r.getExternalDataFromCache(externalDataID, validatorIndex)
```
sorawit, line 155: If user call for `externalDataID = 0` and `validatorIndex = 0`, won't we incorrectly think it's already in the cache?

sorawit (changes requested):
Make sure you write some tests to cover the condition!

chain/owasm/resolver.go
```diff
@@ -6,10 +6,19 @@ import (
 	"github.com/perlin-network/life/exec"
 )
 
+type cache struct {
+	used           bool
```
sorawit: let's call this `isActive`

sorawit (changes requested):
Looks good overall. Just a small nit comment. Let's also update the test to make it more clear.

chain/owasm/resolver_test.go
```diff
@@ -0,0 +1,224 @@
...
+		DefaultMemoryPages:       64,
+		DefaultTableSize:         65536,
+		GasLimit:                 uint64(gasLimit),
+		DisableFloatingPoint:     false,
+		ReturnOnGasLimitExceeded: false,
+	}, resolver, &BandChainGasPolicy{})
+
+	params := [5]int64{}
+
+	// functioncode[69].Numparams is 5
```
taobun: Please explain more what is function code 69?

taobun (changes requested):

chain/owasm/resolver_test.go
```diff
@@ -0,0 +1,265 @@
...
+		if vm.FunctionCode[i].NumParams >= localSize {
+			functionID = i
+			break
+		}
+	}
+	require.NotEqual(t, functionID, -1)
+
+	// Ignite initializes the first call frame.
+	params := [5]int64{}
+	vm.Ignite(functionID, params[:5]...)
```
taobun: If params is a fixed size array that has a length equal to 5, I think we can remove [:5] and uses `params...` instead.

chain/owasm/resolver_test.go
```diff
@@ -0,0 +1,265 @@
...
+	for i := 0; i < len(vm.FunctionCode); i++ {
+		if vm.FunctionCode[i].NumParams >= localSize {
+			functionID = i
+			break
+		}
+	}
+	require.NotEqual(t, functionID, -1)
+
+	// Ignite initializes the first call frame.
+	params := [5]int64{}
```
taobun: ditto

chain/owasm/resolver_test.go
```diff
@@ -0,0 +1,265 @@
...
+	vm.Ignite(functionID, params[:5]...)
+
+	dataSize := resolver.resolveGetCallDataSize(vm)
+	require.Equal(t, dataSize, int64(len(callData)))
```
taobun: Test call data size for?

chain/owasm/resolver_test.go
```diff
@@ -0,0 +1,265 @@
...
+			break
+		}
+	}
+	require.NotEqual(t, functionID, -1)
+
+	// Ignite initializes the first call frame.
+	params := [5]int64{}
+	vm.Ignite(functionID, params[:5]...)
+
+	dataSize := resolver.resolveGetCallDataSize(vm)
```
taobun: Test call data size for?

chain/owasm/resolver_test.go
```diff
@@ -0,0 +1,265 @@
...
+	for i := 0; i < len(vm.FunctionCode); i++ {
+		if vm.FunctionCode[i].NumParams >= localSize {
+			functionID = i
+			break
+		}
+	}
+	require.NotEqual(t, functionID, -1)
+
+	// Ignite initializes the first call frame.
+	params := [5]int64{}
```
taobun: ditto

chain/owasm/resolver_test.go
```diff
@@ -0,0 +1,265 @@
...
+	vm.Ignite(functionID, params[:5]...)
+
+	dataSize := resolver.resolveGetCallDataSize(vm)
```
taobun: ditto

chain/owasm/resolver_test.go
```diff
@@ -0,0 +1,265 @@
...
+	require.Equal(t, dataSize, int64(len(callData)))
+
+	extID := int64(1000)
+	valIndex := int64(1000)
+
+	_, err = resolver.getExternalDataFromCache(extID, valIndex)
+	require.NotNil(t, err)
+
```
taobun, line 362: Can you add test `resolveGetExternalDataSize` and `resolveReadExternalData` that return expected result?

taobun (changes requested):

sorawit (approved):
looks great

taobun (approved):

PR #769, opened 2020-03-09
Generate new private key

taobun:
After merge #764 , It changes the private key after recover from the old mnemonic. This pr change in docker-compose for bandoracled.


sorawit (approved):

evilpeach (approved):
LGTM

PR #770, opened 2020-03-09
scan: build ID type abstraction

### LINKED ISSUE #678: Build abstraction and data types for #ID
This will be reused every where. We should do better than passing int around. Let's design and create custom data types for IDs and function to turn them into components that's clickable.

evilpeach:
Fixes #678

### DIFF
--- a/scan/src/components/TypeID.re
+++ b/scan/src/components/TypeID.re
@@ -15,12 +15,6 @@ let lineHeight =
   | Subtitle => Text.Px(18)
   | Text => Text.Px(16);
 
-let letterSpacing =
-  fun
-  | Title
-  | Subtitle
-  | Text => Text.Em(0.02);
-
 module Styles = {
   open Css;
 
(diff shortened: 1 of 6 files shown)


scan/src/utils/ID.re
```diff
@@ -0,0 +1,26 @@
+type t =
```
sorawit: Let's make custom types for **each** ID type.
```
Module DataSource {
    type t = ID(int)
}

...

type t = 
   | DID(DataSource.t)
   | OID(...)
   | RID(...)
   ...
```

We will be a lot safer in our lives.

sorawit (changes requested):

scan/src/components/TypeID.re
```diff
@@ -0,0 +1,57 @@
...
+  | Subtitle => Text.Lg
+  | Text => Text.Md;
+
+let lineHeight =
+  fun
+  | Title => Text.Px(23)
+  | Subtitle => Text.Px(18)
+  | Text => Text.Px(16);
+
+let letterSpacing =
```
sorawit: Why do we have this function? Seems useless.

sorawit (changes requested):

scan/src/components/TypeID.re
evilpeach: You're right.

taobun (approved):

sorawit (approved):

PR #773, opened 2020-03-10
Add and check parameter that determines how much gas per raw data request

### LINKED ISSUE #578: Add and check parameter that determines how much gas per raw data request
Add `GAS_PER_RAW_DATA_REQUEST` parameter, default to 100,000. After `prepare` is executed, consume this much gas timed the number of raw data requests.

taobun:
fixed #578


chain/x/zoracle/internal/keeper/raw_request.go
```diff
@@ -62,6 +63,10 @@ func (k Keeper) AddNewRawDataRequest(
 		return types.ErrRequestNotFound(types.DefaultCodespace)
 	}
 
+	ctx.GasMeter().ConsumeGas(
+		k.GasPerRawDataRequest(ctx)*uint64(len(request.RequestedValidators)),
```
sorawit: Is it possible to make it overflow here? no?

chain/x/zoracle/internal/types/params.go
```diff
@@ -45,6 +45,9 @@ const (
 	// The maximum size of description length.
 	// Default value 4096
 	DefaultDescriptionLength = int64(4096)
+
+	// Gas cost will be consumed each timed the number of raw data requests and per validators.
```
sorawit: Gas cost per validator for each raw data request.

chain/x/zoracle/genesis.go
```diff
@@ -47,6 +47,7 @@ func InitGenesis(ctx sdk.Context, k Keeper, data GenesisState) []abci.ValidatorU
 	k.SetEndBlockExecuteGasLimit(ctx, data.Params.EndBlockExecuteGasLimit)
 	k.SetMaxNameLength(ctx, data.Params.MaxNameLength)
 	k.SetMaxDescriptionLength(ctx, data.Params.MaxDescriptionLength)
+	k.SetGasPerRawDataRequest(ctx, data.Params.GasPerRawDataRequest)
```
sorawit: Let's call this `GasPerRawDataRequestPerValidator` for extra clarity.

sorawit (changes requested):

chain/x/zoracle/internal/keeper/raw_request.go
taobun: It has a chance if the length of the validators set is too long. We loop consume gas per validator and check overflow in gas consume(cosmos work)?

sorawit: Let's leave it like this for now. I think the chance is minimal.

sorawit (approved):

evilpeach (approved):

PR #786, opened 2020-03-12
chain: Use ErrBadDataValue for bad value errors

### LINKED ISSUE #757: Add `CodeBadDataValue` error code and patch error messages
This error is primarily for data values that violate "Max...." parameters.

sorawit:
Fixes #757


chain/x/zoracle/internal/keeper/raw_request.go
```diff
@@ -22,6 +22,7 @@ func (k Keeper) GetRawDataRequest(
 ) (types.RawDataRequest, sdk.Error) {
 	store := ctx.KVStore(k.storeKey)
 	if !k.CheckRawDataRequestExists(ctx, requestID, externalID) {
+		// TODO: fix error
```
taobun, line 25: Will we not re-consider this error?

sorawit: It will on a different ticket (NotFound error)

taobun (approved):

PR #788, opened 2020-03-12
scan: wire up all datasource

### LINKED ISSUE #684: Wire up all data sources page

evilpeach:
Fixes #684


scan/src/rpc/DataSourceHook.re
```diff
@@ -13,6 +13,7 @@ module DataSource = {
     description: string,
     fee: list(TxHook.Coin.t),
     executable: JsBuffer.t,
+    timestamp: MomentRe.Moment.t,
```
taobun, line 16: Can we use latest revision as a timestamp?

scan/src/rpc/TxHook.re
```diff
@@ -42,6 +42,14 @@ module Coin = {
   let newCoin = (denom, amount) => {denom, amount};
 
   let getDescription = coin => (coin.amount |> Format.fPretty) ++ " " ++ coin.denom;
+
+  let getAmount = coins => {
+    let coinOpt = coins->Belt_List.get(0);
```
taobun, line 70: Assume that it has only one element?

evilpeach: Its name is misleading. It should be named `getFeeAmount`. Currently, I think we don't need to support multiple fees.

scan/src/pages/DataSourceHomePage.re
```diff
@@ -23,10 +23,23 @@ module Styles = {
...
+      height(`px(200)),
+      boxShadow(Shadow.box(~x=`px(0), ~y=`px(2), ~blur=`px(2), Css.rgba(0, 0, 0, 0.05))),
+      backgroundColor(white),
+    ]);
 };
 
 [@react.component]
 let make = () => {
+  let dataSourcesOpt = DataSourceHook.getList();
+  Js.Console.log(dataSourcesOpt);
```
taobun: Remove log

taobun (changes requested):

scan/src/pages/DataSourceHomePage.re
```diff
@@ -106,57 +122,32 @@ let make = () => {
...
+               <div className=Styles.fullWidth>
+                 <Row>
+                   <Col> <HSpacing size=Spacing.xl /> </Col>
+                   <Col size=0.5>
+                     <TElement elementType={TElement.DataSource(ID.DataSource.ID(id), name)} />
+                   </Col>
+                   <Col size=0.5> <TElement elementType={timestamp->TElement.Timestamp} /> </Col>
+                   <Col size=1.> <TElement elementType={owner->TElement.Address} /> </Col>
+                   <Col size=0.4>
+                     <TElement elementType={fee->TxHook.Coin.getFeeAmount->TElement.Fee} />
```
taobun, line 139: Expect to reuse Fee type? or float type?

taobun: image

scan/src/rpc/DataSourceHook.re
evilpeach: I think we should get it from the end point.

taobun (approved):

prin-r (approved):

PR #800, opened 2020-03-13
scan: update Timeago

evilpeach:
Support spacing, prefix, suffix in `timeagos`

### DIFF
--- a/scan/src/components/TimeAgos.re
+++ b/scan/src/components/TimeAgos.re
@@ -1,4 +1,8 @@
-open Css;
+module Styles = {
+  open Css;
+
+  let timeContainer = style([display(`inlineFlex)]);
+};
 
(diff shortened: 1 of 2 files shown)


taobun (dismissed):

scan/src/components/TimeAgos.re
```diff
@@ -1,3 +1,5 @@
+open Css;
```
sorawit: Do not open Css on a file level. It will hurt you as Css contains so many common names. Either declare `Styles` module and declare styles in there, or use `Css.` everywhere. Your choice.

sorawit (changes requested):

scan/src/components/TimeAgos.re
evilpeach: You're right

sorawit (approved):

Benzbeeb (approved):

PR #819, opened 2020-03-13
show messages

prin-r:
show messages for txs page

### DIFF
--- a/scan/src/components/TxsTable.re
+++ b/scan/src/components/TxsTable.re
@@ -1,10 +1,17 @@
 module Styles = {
   open Css;
+  let hScale = 20;
   let fullWidth = style([width(`percent(100.0)), display(`flex)]);
   let hashContainer = style([maxWidth(`px(140))]);
   let statusContainer =
     style([maxWidth(`px(95)), display(`flex), flexDirection(`row), alignItems(`center)]);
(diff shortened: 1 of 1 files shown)


evilpeach (approved):

scan/src/components/TxsTable.re
```diff
@@ -50,42 +57,66 @@ let make = (~txs: list(TxHook.Tx.t)) => {
...
              <HSpacing size={`px(20)} />
-             <Col size=1.67> <TxLink txHash=hash width=140 /> </Col>
-             <Col size=0.88> <TypeID.Block id={ID.Block.ID(blockHeight)} /> </Col>
+             <Col size=1.67>
+               <div className={Styles.heightByMsgsNum(numMsgs, 0)}>
+                 <TxLink txHash=hash width=140 />
+               </div>
+             </Col>
+             <Col size=0.88>
+               <div className={Styles.heightByMsgsNum(numMsgs, -4)}>
```
taobun, line 70: Why it must be -4 and if more msg this margin would be scale follow a number of messages?

prin-r: The good way to do this would need more time. So right now will fix it by using magic number. And we might come back to fix it if needed.

taobun (approved):

PR #822, opened 2020-03-13
Make rpc url

### LINKED ISSUE #751: Make RPC url of scan configurable
Should be through env variable. Also work with Swit to update render.com env variable.

Here's some pointer: https://parceljs.org/env.html

prin-r:
Fix #751

### DIFF
--- a/scan/package.json
+++ b/scan/package.json
@@ -6,7 +6,7 @@
     "parcel-bundler": "^1.12.4"
   },
   "scripts": {
-    "build": "bsb -make-world && parcel build index.html",
+    "build": "rm -rf dist/ && bsb -make-world && parcel build index.html --no-source-maps --no-cache",
     "start": "bsb -make-world -w",
     "test": "yarn build && yarn jest --coverage"
   },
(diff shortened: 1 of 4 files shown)


scan/package.json
```diff
@@ -6,7 +6,7 @@
     "parcel-bundler": "^1.12.4"
   },
   "scripts": {
-    "build": "bsb -make-world && parcel build index.html",
+    "build": "rm -rf dist/ && bsb -make-world && parcel build index.html --no-source-maps --no-cache",
```
sorawit, line 9: Why is this change necessary?

prin-r: --no-cache is necessary because sometime it will use the env from cache. So we just remove the cache.
"rm -rf dist/" and "--no-source-maps" are irrelevant to this pr, I just add them to reduce redundant build in dist folder and reduce the size of the bundle.

sorawit: Sounds good!

sorawit (approved):

sorawit:
I did set environment variable in render.com to have `RPC_URL=https://d3n.bandprotocol.com/` and it seems to be working just fine!

evilpeach (approved):

PR #851, opened 2020-03-15
lambda: Add lambda module

sorawit:
The module can be uploaded to AWS lambda using Python3.8 runtime for data source execution. Note that all dependency binaries and libraries are also included here, as they need to be built with Amazon Linux.


taobun (approved):

sorawit:
@prin-r are you good with this?

taobun:
@prin-r please review this pr

prin-r (approved):

PR #864, opened 2020-03-16
chain: Add events for simple operations

sorawit:
This includes CreateDataSource, EditDataSource, CreateOracleScript, and EditOracleScript

### DIFF
--- a/chain/x/zoracle/internal/keeper/data_source.go
+++ b/chain/x/zoracle/internal/keeper/data_source.go
@@ -13,25 +13,28 @@ func (k Keeper) SetDataSource(ctx sdk.Context, id types.DataSourceID, dataSource
 }
 
 // AddDataSource adds the given data source to the storage.
-func (k Keeper) AddDataSource(ctx sdk.Context, owner sdk.AccAddress, name string, description string, fee sdk.Coins, executable []byte) sdk.Error {
+func (k Keeper) AddDataSource(
+	ctx sdk.Context, owner sdk.AccAddress, name string, description string,
+	fee sdk.Coins, executable []byte,
+) (types.DataSourceID, sdk.Error) {
 	newDataSourceID := k.GetNextDataSourceID(ctx)
 
 	if len(executable) > int(k.MaxDataSourceExecutableSize(ctx)) {
-		return types.ErrBadDataValue(
+		return 0, types.ErrBadDataValue(
 			"AddDataSource: Executable size (%d) exceeds the maximum size (%d).",
 			len(executable),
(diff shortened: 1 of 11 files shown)


chain/x/zoracle/internal/keeper/data_source.go
```diff
@@ -13,25 +13,28 @@ func (k Keeper) SetDataSource(ctx sdk.Context, id types.DataSourceID, dataSource
 }
 
 // AddDataSource adds the given data source to the storage.
-func (k Keeper) AddDataSource(ctx sdk.Context, owner sdk.AccAddress, name string, description string, fee sdk.Coins, executable []byte) sdk.Error {
+func (k Keeper) AddDataSource(
```
taobun, line 16: Please just add a simple test for adding data source and check data source id that return from the function is correct.

chain/x/zoracle/internal/keeper/oracle_script.go
```diff
@@ -13,25 +13,27 @@ func (k Keeper) SetOracleScript(ctx sdk.Context, id types.OracleScriptID, oracle
 }
 
 // AddOracleScript adds the given oracle script to the storage.
-func (k Keeper) AddOracleScript(ctx sdk.Context, owner sdk.AccAddress, name string, description string, code []byte) sdk.Error {
+func (k Keeper) AddOracleScript(
```
taobun, line 16: ditto

taobun (changes requested):

chain/x/zoracle/internal/keeper/data_source.go
taobun: Or you can test on handler that emits the event correctly,

chain/x/zoracle/internal/types/events.go
```diff
@@ -2,8 +2,14 @@ package types
...
-	EventTypeReport       = "report"
+	EventTypeCreateDataSource   = "create_data_source"
+	EventTypeEditDataSource     = "edit_data_source"
+	EventTypeCreateOracleScript = "create_oracle_script"
+	EventTypeEditOracleScript   = "edit_oracle_script"
+	EventTypeRequest            = "request"
+	EventTypeReport             = "report"
+
+	AttributeKeyID        = "id"
 	AttributeKeyRequestID = "id"
```
taobun: Remove this?

sorawit: No it shouldn't be removed because it's also use in Report event. I rename the value to  "request_id" though for clarity.

taobun (approved):

PR #866, opened 2020-03-16
chain: Refund transaction fee if success with all report messages

sorawit:


prin-r (commented):
This pr was tested by comparing report tx with send tx.
Both txs were sent by validator1 with fee = 500,000 uband.
Validator1 lose nothing after sending report tx but lose 500,888 uband for sending tx (fee 500,00, amount 888).

[reportTx.txt](https://github.com/bandprotocol/bandchain/files/4338285/reportTx.txt)
[sendTx.txt](https://github.com/bandprotocol/bandchain/files/4338287/sendTx.txt)

chain/app/app.go
```diff
@@ -361,20 +343,56 @@ func NewBandApp(
...
+					isAllReportTxs = false
+				}
+			}
+			if isAllReportTxs {
+				acc := app.accountKeeper.GetAccount(app.deliverContext, stdTx.GetSigners()[0])
+				if acc != nil {
+					err := app.supplyKeeper.SendCoinsFromModuleToAccount(
+						app.deliverContext,
+						auth.FeeCollectorName,
+						acc.GetAddress(),
```
taobun: Why do we need to get an account from the account keeper? Using stdTx.GetSigners()[0] as to address from send coin is not enough?

chain/app/app.go
```diff
@@ -361,20 +343,56 @@ func NewBandApp(
...
+						stdTx.Fee.Amount,
```
taobun: wow feels good

chain/app/app.go
sorawit: I'm not sure, I just strictly follow what Cosmos ante code is doing. Can you help double check? Cosmos `x/auth/ante.go` specifically calls `GetSignerAcc` to get account from address, then call `DeductFees` on the account, though `DeductFees` just calls `SendCoinsFromModuleToAccount` on the acccount's address.

sorawit: Is it possible that the account is removed somehow and somehow reach here without an account on that address?

chain/app/app.go
```diff
@@ -123,6 +122,11 @@ type bandApp struct {
 	paramsKeeper   params.Keeper
 	zoracleKeeper  zoracle.Keeper
 
+	// Decoder for unmarshaling []byte into sdk.Tx
+	txDecoder sdk.TxDecoder
+	// Deliver Context that is set during BeingBlock and unset during EndBlock; primarily for gas refund
+	deliverContext sdk.Context
```
taobun, line 128: Make it to public or create getter function to get it

chain/app/app.go
```diff
@@ -361,20 +343,54 @@ func NewBandApp(
...
+		// Refund 100% of gas fee for any successful transaction that only contains MsgReportData
+		tx, err := app.txDecoder(req.Tx)
+		if err != nil { // Should never happen because BaseApp.DeliverTx succeeds
+			panic(err)
+		}
+		isAllReportTxs := true
+		if stdTx, ok := tx.(auth.StdTx); ok {
+			for _, msg := range tx.GetMsgs() {
+				if _, ok := msg.(zoracle.MsgReportData); !ok {
+					isAllReportTxs = false
```
taobun, line 373: break loop?

taobun (changes requested):

chain/app/app.go
sorawit: That will be on a separate PR to make everything including all keepers accessible.

chain/app/app.go
sorawit: Right. thanks!

taobun (approved):

PR #945, opened 2020-03-20
Update scan url route scheme

### LINKED ISSUE #821: Update scan URL route scheme
- `/validators`
- `/sources`
- `/scripts`
- `/blocks`
- `/txs`
- `/DXXXX`, `/OXXXX`, `/BXXXX`, `/RXXXX`
- `/bandxxxxxxx`
- `/bandvaloperxxxxx`
- `/tx/DEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEF`

prin-r:
Fix #821

### DIFF
--- a/scan/src/bindings/Bech32.re
+++ b/scan/src/bindings/Bech32.re
@@ -4,6 +4,26 @@ type decoded_t = {
   words: array(int),
 };
 
+let safeDecode: string => (decoded_t, bool) = [%bs.raw
+  {|
+function(str) {
+  const bech32 = require("bech32");
+  try {
+    const result = bech32.decode(str);
+    return [result,true];
+  } catch(_) {
+    return [{prefix:"",words:[]}, false];
+  }
+}
+  |}
+];
+
+let decodeOpt = str =>
+  switch (str->safeDecode) {
+  | (result, true) => Some(result)
+  | _ => None
+  };
+
 [@bs.module "bech32"] [@bs.val] external fromWords: array(int) => array(int) = "fromWords";
 [@bs.module "bech32"] [@bs.val] external toWords: array(int) => array(int) = "toWords";
 
--- a/scan/src/bindings/Bech32.re
+++ b/scan/src/bindings/Bech32.re
@@ -4,28 +4,14 @@ type decoded_t = {
   words: array(int),
 };
 
-let safeDecode: string => (decoded_t, bool) = [%bs.raw
-  {|
-function(str) {
-  const bech32 = require("bech32");
-  try {
-    const result = bech32.decode(str);
-    return [result,true];
-  } catch(_) {
-    return [{prefix:"",words:[]}, false];
-  }
-}
-  |}
-];
-
-let decodeOpt = str =>
-  switch (str->safeDecode) {
-  | (result, true) => Some(result)
-  | _ => None
-  };
-
 [@bs.module "bech32"] [@bs.val] external fromWords: array(int) => array(int) = "fromWords";
 [@bs.module "bech32"] [@bs.val] external toWords: array(int) => array(int) = "toWords";
 
 [@bs.module "bech32"] [@bs.val] external decode: string => decoded_t = "decode";
 [@bs.module "bech32"] [@bs.val] external encode: (string, array(int)) => string = "encode";
+
+let decodeOpt = str =>
+  switch (str->decode) {
+  | result => Some(result)
+  | exception _ => None
+  };
--- a/scan/src/utils/Address.re
+++ b/scan/src/utils/Address.re
@@ -5,6 +5,13 @@ let fromBech32 = bech32str => {
   Address(bech32str->Bech32.decode->Bech32.wordsGet->Bech32.fromWords->JsBuffer.arrayToHex);
 };
 
+let fromBech32Opt = bech32str =>
+  switch (bech32str->Bech32.decodeOpt) {
+  | Some(decoded) =>
+    Some(Address(decoded->Bech32.wordsGet->Bech32.fromWords->JsBuffer.arrayToHex))
+  | _ => None
+  };
+
 let fromHex = hexstr => Address(hexstr->HexUtils.normalizeHexString);
 
 let toHex = (~with0x=false) =>
--- a/scan/src/utils/Address.re
+++ b/scan/src/utils/Address.re
@@ -6,10 +6,9 @@ let fromBech32 = bech32str => {
 };
 
 let fromBech32Opt = bech32str =>
-  switch (bech32str->Bech32.decodeOpt) {
-  | Some(decoded) =>
-    Some(Address(decoded->Bech32.wordsGet->Bech32.fromWords->JsBuffer.arrayToHex))
-  | _ => None
+  switch (bech32str |> fromBech32) {
+  | result => Some(result)
+  | exception _ => None
   };
 
 let fromHex = hexstr => Address(hexstr->HexUtils.normalizeHexString);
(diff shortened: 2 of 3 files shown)


sorawit (changes requested):
Let's use regex pattern match & extraction here. Make definition of each route clearer.

scan/src/utils/Route.re
```diff
@@ -39,87 +39,89 @@ type t =
...
+    | (["O", ...rest], _) => OracleScriptIndexPage(rest |> chars_to_int, OracleScriptExecute)
+    | (["D", ...rest], "code") => DataSourceIndexPage(rest |> chars_to_int, DataSourceCode)
+    | (["D", ...rest], "requests") =>
+      DataSourceIndexPage(rest |> chars_to_int, DataSourceRequests)
+    | (["D", ...rest], "revisions") =>
+      DataSourceIndexPage(rest |> chars_to_int, DataSourceRevisions)
+    | (["D", ...rest], _) => DataSourceIndexPage(rest |> chars_to_int, DataSourceExecute)
+    | (["R", ...rest], "proof") => RequestIndexPage(rest |> chars_to_int, RequestProof)
+    | (["R", ...rest], _) => RequestIndexPage(rest |> chars_to_int, RequestReportStatus)
+    | (["b", "a", "n", "d", "v", "a", "l", "o", "p", "e", "r", ..._], _) =>
```
taobun: Seem weird use startwith like search will be better?  @sorawit

sorawit: My recommendation is that we should explore using regex matching / group extraction. @prin-r already started looking into this yesterday.

scan/src/bindings/Bech32.re
```diff
@@ -4,6 +4,26 @@ type decoded_t = {
   words: array(int),
 };
 
+let safeDecode: string => (decoded_t, bool) = [%bs.raw
```
sorawit: Good try. But I don't think you need to go bs_raw here. You can catch exception in ReasonML. https://reasonml.github.io/docs/en/exception.

scan/src/utils/Route.re
```diff
@@ -39,87 +39,165 @@ type t =
   | ValidatorHomePage
   | ValidatorIndexPage(Address.t, validator_tab_t);
 
+let regexes = [
+  ("^$" |> Js.Re.fromString, _ => HomePage),
```
sorawit: Can we do `|> Js.Re.fromString` only once in `keepMap`?

scan/src/utils/Route.re
```diff
@@ -39,87 +39,165 @@ type t =
...
-    ValidatorIndexPage(address |> Address.fromBech32, Reports)
-  | (["validator", address], _) =>
-    ValidatorIndexPage(address |> Address.fromBech32, ProposedBlocks)
-  | ([], _) => HomePage
-  | (_, _) => NotFound
+  switch (
+    regexes
+    ->Belt_List.keepMap(((regex, keysToRoute)) =>
+        url.path
+        ->Belt_List.get(0)
```
sorawit: Can also do `Belt_List.head`

scan/src/utils/Route.re
```diff
@@ -39,87 +39,165 @@ type t =
...
+        ->Belt_Option.getWithDefault("")
+        ->Js.Re.exec_(regex, _)
```
sorawit: Here you can do `|> Js.Re.fromString`

scan/src/utils/Route.re
```diff
@@ -39,87 +39,165 @@ type t =
...
+      )
+    ->Belt.List.head
+  ) {
+  | Some((result, keysToRoute)) =>
+    keysToRoute(
+      result
+      ->Js.Re.captures
+      ->Belt_Array.keepMap(Js.toOption)
+      ->Belt_List.fromArray
+      ->Belt_List.concat(url.path->Belt_List.drop(1)->Belt_Option.getWithDefault([]))
```
sorawit: This is too much of a confusion. You are concating what you get from regex extractions with other parts of path. Don't do that.

scan/src/utils/Route.re
```diff
@@ -39,87 +39,165 @@ type t =
...
+      switch (keys->Belt_List.get(1)->Belt_Option.getWithDefault("")->int_of_string_opt) {
+      | Some(height) => BlockIndexPage(height)
+      | None => NotFound
+      };
+    },
+  ),
+  (
+    "D([0-9]+)$" |> Js.Re.fromString,
+    keys => {
+      switch (
```
sorawit: IMO. Makes more sense to make this return
```
DataSourceIndexPage(
    keys->Belt_List.getExn(1)->int_of_string, 
    dataSourceTab(keys->Belt_List.getExn(2)),
)
```
and define `dataSourceTab` function separately.

sorawit (changes requested):

sorawit:
Please try to make each function / data type have good standalone meaning.

scan/src/utils/Route.re
prin-r: What should we do if keys->Belt_List.getExn(2) not in ["","code","requests","revisions"]

prin-r: I currently solve this by using
    keys =>
      dataSourceTab(keys->Belt_List.getExn(2))
      ->Belt_Option.mapWithDefault(NotFound, tab =>
          DataSourceIndexPage(keys->Belt_List.getExn(1)->int_of_string, tab)
        ),

scan/src/utils/Route.re
prin-r: This is happen because "tx/abcd...".
Other paths don't have "/". So url.path only contain 1 element such as "D12", "bandvfsettg...".
But for tx, its path will contain 2 elements (["tx", "abcd..."]).

evilpeach (dismissed):

evilpeach:
@prin-r Please fix the conflict

scan/src/utils/Address.re
```diff
@@ -5,6 +5,13 @@ let fromBech32 = bech32str => {
   Address(bech32str->Bech32.decode->Bech32.wordsGet->Bech32.fromWords->JsBuffer.arrayToHex);
 };
 
+let fromBech32Opt = bech32str =>
+  switch (bech32str->Bech32.decodeOpt) {
```
sorawit: Should we just do
```
switch (bech32str |> fromBech32) {
| result => Some(result)
| exception _ => None
}
```
then get rid of decodeOpt?

scan/src/utils/Route.re
sorawit: No. let's split into separate variables

sorawit (changes requested):

scan/src/utils/Route.re
```diff
@@ -39,87 +79,148 @@ type t =
   | ValidatorHomePage
   | ValidatorIndexPage(Address.t, validator_tab_t);
 
+type url_parts_t = {
+  first: list(string),
```
sorawit, line 83: Why is this called `first`? Shouldn't it be called `captures`?

scan/src/bindings/Bech32.re
```diff
@@ -9,3 +9,9 @@ type decoded_t = {
 
 [@bs.module "bech32"] [@bs.val] external decode: string => decoded_t = "decode";
 [@bs.module "bech32"] [@bs.val] external encode: (string, array(int)) => string = "encode";
+
+let decodeOpt = str =>
```
sorawit, line 13: This is not used, but you can keep it if you want.

sorawit (changes requested):

sorawit:
Far outdated by now.

PR #948, opened 2020-03-21
chain: Make db handle each transaction instead of event

sorawit:


chain/app/db_app.go
```diff
@@ -48,14 +51,24 @@ func (app *dbBandApp) DeliverTx(req abci.RequestDeliverTx) (res abci.ResponseDel
...
-	if res.IsOK() && lastProcessHeight+1 == app.DeliverContext.BlockHeight() {
-		for _, event := range res.Events {
-			kvMap := make(map[string]string)
-			for _, kv := range event.Attributes {
-				kvMap[string(kv.GetKey())] = string(kv.GetValue())
-			}
-			app.dbBand.HandleEvent(event.Type, kvMap)
-		}
+	if lastProcessHeight+1 == app.DeliverContext.BlockHeight() {
+		return res
```
sorawit, line 55: This is not correct, as @taobun pointed out.

chain/db/db.go
```diff
@@ -50,18 +54,37 @@ func (b *BandDB) RollBack() {
 	b.tx = nil
 }
 
-func (b *BandDB) HandleEvent(eventName string, attributes map[string]string) {
-	switch eventName {
+func (b *BandDB) HandleTransaction(tx auth.StdTx, txHash []byte, logs sdk.ABCIMessageLogs) {
```
sorawit, line 57: We may also need other information like block time, block height. Might even be better to just pass Context here?

sorawit:
@taobun Please take this forward

chain/db/db.go
taobun: We need info about that Send context here or save context in bandDB state?

sorawit:
Close in favor of #957

PR #957, opened 2020-03-22
Implement caching layer for data source information

### LINKED ISSUE #916: Implement caching layer for data source information
The data source table should contain
- ID (pk)
- name
- description
- owner
- executable
- request fee
- last update

Also add data source revision table that should contain
- ID (pk)
- Revision number (pk)
- Name
- timestamp
- block height
- tx hash

taobun:
fixed #916


chain/app/db_app.go
```diff
@@ -25,6 +29,7 @@ func NewDBBandApp(
 ) *dbBandApp {
 	app := NewBandApp(logger, db, traceStore, loadLatest, invCheckPeriod, baseAppOptions...)
 
+	dbBand.ZoracleKeeper = app.ZoracleKeeper
```
sorawit, line 32: Oh god.... Do we really need to do this for all keepers?

chain/db/data_source.go
```diff
@@ -0,0 +1,105 @@
...
+		return errors.New("handleMsgCreateDataSource: cannot find data source id")
+	}
+	id, err := strconv.ParseInt(rawID, 10, 64)
+	if err != nil {
+		return err
+	}
+	return b.AddDataSource(
+		id, msg.Name, msg.Description, msg.Owner, msg.Fee, msg.Executable,
+		b.ctx.BlockTime(), b.ctx.BlockHeight(), txHash,
+	)
```
sorawit, line 80: I think we also add to revision here.

chain/db/db.go
```diff
@@ -33,6 +42,13 @@ func NewDB(dialect, path string, metadata map[string]string) (*BandDB, error) {
 		"RESTRICT",
 	)
 
+	db.Model(&DataSourceRevision{}).AddForeignKey(
+		"data_source_id",
+		"data_sources(id)",
+		"CASCADE",
```
sorawit: Why don't we do RESTRICT? I think it's safer that way.

chain/db/data_source.go
```diff
@@ -0,0 +1,105 @@
...
+		msg.Owner, msg.Fee, msg.Executable, b.ctx.BlockTime(),
+	)
+
+	err := b.tx.Save(&dataSource).Error
+	if err != nil {
+		return err
+	}
+
+	return b.tx.Create(&DataSourceRevision{
+		DataSourceID: int64(msg.DataSourceID),
```
sorawit, line 99: Why do we ignore `RevisionNumber`?

sorawit (changes requested):

chain/db/data_source.go
taobun: AddDataSource is a function to create new data source row with revision already.

chain/db/data_source.go
taobun: I use auto increment for this field

chain/db/db.go
taobun: In my opinion, this table will never be deleted by logic, but it's okay to use RESTRICT

chain/app/db_app.go
taobun: :(

taobun: If we don't work like this how we send keeper to handleMsg

chain/app/db_app.go
```diff
@@ -101,6 +138,7 @@ func (app *dbBandApp) BeginBlock(req abci.RequestBeginBlock) (res abci.ResponseB
 	res = app.bandApp.BeginBlock(req)
 	// Begin transaction
 	app.dbBand.BeginTransaction()
+	app.dbBand.SetContext(app.DeliverContext)
```
sorawit, line 141: Let's also clear this after EndBlock. Also should we make it readonly. Like creating a cache on top and discard after.

sorawit (changes requested):

chain/db/table.go
```diff
@@ -27,3 +27,22 @@ type ValidatorVote struct {
 	BlockHeight      int64  `gorm:"primary_key;auto_increment:false"`
 	Voted            bool
 }
+
+type DataSource struct {
+	ID          int64 `gorm:"primary_key;auto_increment:false"`
+	Name        string
```
sorawit: Let's decorate everything that's not nullable `not null`. We will have better life with GraphQL that way, right @evilpeach ?

sorawit (changes requested):

chain/db/table.go
evilpeach: Absolutely

taobun: Already did

sorawit (dismissed):

sorawit (approved):

evilpeach (approved):

PR #962, opened 2020-03-24
Do not keep context and keeper inside of exec_env

### LINKED ISSUE #947: Do not keep context and keeper inside of exec_env
To prevent the gas issue during Wasm execution. 
For prepare: We simply collect raw data requests to memory. Then save those raw requests later.
For execute: We preemptively load all reports from all validators in memory first.

Benzbeeb:
Fixed #947


chain/x/zoracle/exec_env.go
```diff
@@ -65,19 +86,55 @@ func (env *ExecutionEnvironment) GetValidatorAddress(validatorIndex int64) ([]by
...
+		types.NewRawDataRequest(types.DataSourceID(dataSourceID), calldata),
+	))
+}
+
+func (env *ExecutionEnvironment) LoadAllRawDataReports(
+	ctx sdk.Context,
+	keeper Keeper,
+) sdk.Error {
+
+	tmp := []types.ReportWithValidator{}
```
taobun: No need to append to the slice,  update map in this loop

chain/x/zoracle/exec_env.go
```diff
@@ -88,14 +145,15 @@ func (env *ExecutionEnvironment) GetExternalData(
...
+	rawDataReport, ok := env.rawDataReports[key]
+
+	if !ok {
+		errMsg := fmt.Sprintf("Unable to find raw data report with request ID %d external ID %d from %s\n", int64(env.requestID), externalDataID, validatorAddress.String())
+
+		return nil, 0, errors.New(errMsg)
 	}
-	return rawReport.Data, rawReport.ExitCode, nil
+
+	return rawDataReport.Data, env.rawDataReports[key].ExitCode, nil
```
taobun: Use rawDataReport.ExitCode?

taobun (changes requested):

chain/x/zoracle/exec_env.go
```diff
@@ -65,19 +86,47 @@ func (env *ExecutionEnvironment) GetValidatorAddress(validatorIndex int64) ([]by
...
-) error {
-	return env.keeper.AddNewRawDataRequest(env.ctx, env.requestID, types.ExternalID(externalDataID), types.DataSourceID(dataSourceID), calldata)
+) {
+	env.rawDataRequests = append(env.rawDataRequests, types.NewRawDataRequestWithExternalID(
+		types.ExternalID(externalDataID),
+		types.NewRawDataRequest(types.DataSourceID(dataSourceID), calldata),
+	))
+}
+
+func (env *ExecutionEnvironment) LoadAllRawDataReports(
```
sorawit: Move this to be after `SaveRawDataRequests` and rename to `LoadRawDataReports`

chain/x/zoracle/exec_env.go
```diff
@@ -65,19 +86,47 @@ func (env *ExecutionEnvironment) GetValidatorAddress(validatorIndex int64) ([]by
...
 }
 
 func (env *ExecutionEnvironment) RequestExternalData(
 	dataSourceID int64,
 	externalDataID int64,
 	calldata []byte,
-) error {
-	return env.keeper.AddNewRawDataRequest(env.ctx, env.requestID, types.ExternalID(externalDataID), types.DataSourceID(dataSourceID), calldata)
+) {
+	env.rawDataRequests = append(env.rawDataRequests, types.NewRawDataRequestWithExternalID(
```
sorawit, line 131: Let's make sure there is an upper bound on the slice length. Load max raw data request count into here. Then if the size already exceeds that param, just ignore and not push it to the slice.

sorawit (changes requested):

sorawit:
@taobun @Benzbeeb With this diff, I don't think caching is necessary anymore, right?

chain/x/zoracle/exec_env.go
```diff
@@ -65,19 +111,25 @@ func (env *ExecutionEnvironment) GetValidatorAddress(validatorIndex int64) ([]by
...
 
 func (env *ExecutionEnvironment) RequestExternalData(
 	dataSourceID int64,
 	externalDataID int64,
 	calldata []byte,
-) error {
-	return env.keeper.AddNewRawDataRequest(env.ctx, env.requestID, types.ExternalID(externalDataID), types.DataSourceID(dataSourceID), calldata)
+) {
+	if int64(len(calldata)) <= env.maxCalldataSize {
+		env.rawDataRequests = append(env.rawDataRequests, types.NewRawDataRequestWithExternalID(
```
sorawit: Let make this a no-op when the size of `rawDataRequests` already exceeds max raw data request count, so Wasm cannot flood us with shit tons of raw requests.

sorawit (changes requested):

chain/owasm/resolver.go
```diff
@@ -132,26 +122,13 @@ func (r *resolver) resolveRequestExternalData(vm *exec.VirtualMachine) int64 {
...
 	copy(data, vm.Memory[dataOffset:dataOffset+dataLength])
-	err := r.env.RequestExternalData(dataSourceID, externalDataID, data)
-	if err != nil {
-		return -1
-	}
+	r.env.RequestExternalData(dataSourceID, externalDataID, data)
 	return 0
 }
 
 func (r *resolver) getExternalDataFromCache(externalDataID int64, validatorIndex int64) ([]byte, uint8, error) {
```
taobun: Remove it

taobun (changes requested):

chain/owasm/resolver.go
```diff
@@ -132,34 +122,21 @@ func (r *resolver) resolveRequestExternalData(vm *exec.VirtualMachine) int64 {
...
-	externalData, statusCode, err := r.env.GetExternalData(externalDataID, validatorIndex)
-	r.cachedata = cache{
-		externalDataID: externalDataID,
-		validatorIndex: validatorIndex,
-		data:           externalData,
-		statusCode:     statusCode,
-		err:            err,
-		isActive:       true,
-	}
+// func (r *resolver) getExternalDataFromCache(externalDataID int64, validatorIndex int64) ([]byte, uint8, error) {
```
taobun: Not commend and delete it

taobun (changes requested):

chain/x/zoracle/exec_env.go
```diff
@@ -63,20 +114,18 @@ func (env *ExecutionEnvironment) GetValidatorAddress(validatorIndex int64) ([]by
...
-}
-
 func (env *ExecutionEnvironment) RequestExternalData(
 	dataSourceID int64,
 	externalDataID int64,
 	calldata []byte,
-) error {
-	return env.keeper.AddNewRawDataRequest(env.ctx, env.requestID, types.ExternalID(externalDataID), types.DataSourceID(dataSourceID), calldata)
+) {
+	if int64(len(calldata)) <= env.maxCalldataSize {
```
sorawit: You should ignore only when  `int64(len(calldata)) > env.maxCalldataSize + 1`. That way the function is guarantee to fail down the road.

sorawit: Talk to me if you don't understand what's going on.

sorawit: This does not fix what I comment at all. Do not just click resolve without actually dealing with it.

sorawit (changes requested):

chain/x/zoracle/exec_env.go
```diff
@@ -63,20 +116,18 @@ func (env *ExecutionEnvironment) GetValidatorAddress(validatorIndex int64) ([]by
...
-}
-
 func (env *ExecutionEnvironment) RequestExternalData(
 	dataSourceID int64,
 	externalDataID int64,
 	calldata []byte,
-) error {
-	return env.keeper.AddNewRawDataRequest(env.ctx, env.requestID, types.ExternalID(externalDataID), types.DataSourceID(dataSourceID), calldata)
+) {
+	if int64(len(calldata)) <= env.maxCalldataSize && len(env.rawDataRequests) < int(env.maxDataRequestCount) {
```
sorawit: `int64(len(...)) <= env.maxData...`

sorawit: Also add comment explaining the logic.

sorawit (changes requested):

chain/x/zoracle/exec_env.go
```diff
@@ -63,20 +116,19 @@ func (env *ExecutionEnvironment) GetValidatorAddress(validatorIndex int64) ([]by
...
-}
-
 func (env *ExecutionEnvironment) RequestExternalData(
 	dataSourceID int64,
 	externalDataID int64,
 	calldata []byte,
-) error {
-	return env.keeper.AddNewRawDataRequest(env.ctx, env.requestID, types.ExternalID(externalDataID), types.DataSourceID(dataSourceID), calldata)
+) {
+	// when size of env.rawDataRequests == env.maxRawDataRequest that mean it exceeds the number of request
```
sorawit: // We can safely skip appending rawDataRequests if its size already surpasses maxRawDataRequestCount threshold since it will fail the transaction regardless.

sorawit (changes requested):

chain/x/zoracle/exec_env.go
```diff
@@ -63,20 +116,19 @@ func (env *ExecutionEnvironment) GetValidatorAddress(validatorIndex int64) ([]by
...
-
 func (env *ExecutionEnvironment) RequestExternalData(
 	dataSourceID int64,
 	externalDataID int64,
 	calldata []byte,
-) error {
-	return env.keeper.AddNewRawDataRequest(env.ctx, env.requestID, types.ExternalID(externalDataID), types.DataSourceID(dataSourceID), calldata)
+) {
+	// We can safely skip appending rawDataRequests if its size already surpasses maxRawDataRequestCount threshold since it will fail the transaction regardless.
+	if int64(len(calldata)) <= env.maxCalldataSize && int64(len(env.rawDataRequests)) <= env.maxRawDataRequestCount {
```
sorawit: if `int64(len(calldata)) > env.maxCalldataSize` happens, we should just return -1 back to the requester, I think. Same for when `int64(len(env.rawDataRequests)) >= env.maxRawDataRequestCount`

sorawit (changes requested):

chain/x/zoracle/exec_env.go
```diff
@@ -63,20 +116,20 @@ func (env *ExecutionEnvironment) GetValidatorAddress(validatorIndex int64) ([]by
...
-	return int64(env.keeper.GetParam(env.ctx, KeyMaxCalldataSize))
-}
-
 func (env *ExecutionEnvironment) RequestExternalData(
 	dataSourceID int64,
 	externalDataID int64,
 	calldata []byte,
-) error {
-	return env.keeper.AddNewRawDataRequest(env.ctx, env.requestID, types.ExternalID(externalDataID), types.DataSourceID(dataSourceID), calldata)
+) int64 {
```
sorawit: This should return error to be consistent with other functions (see GetExternalData as example). In `resolver` you return -1 if error is not nil.

sorawit (changes requested):

chain/owasm/mock.go
```diff
@@ -60,7 +60,6 @@ func (m *mockExecutionEnvironment) RequestExternalData(
 	calldata []byte,
 ) error {
 	// TODO: Figure out how to mock this elegantly.
-	fmt.Printf("RequestExternalData: DataSourceID = %d, ExternalDataID = %d\n", dataSourceID, externalDataID)
```
sorawit: Why did we remove this?

sorawit (changes requested):

chain/x/zoracle/exec_env.go
```diff
@@ -87,14 +143,13 @@ func (env *ExecutionEnvironment) GetExternalData(
...
-		validatorAddress,
-	)
-	if err != nil {
-		return nil, 0, err
+	key := string(types.RawDataReportStoreKey(env.requestID, types.ExternalID(externalDataID), validatorAddress))
+
+	rawDataReport, ok := env.rawDataReports[key]
+
+	if !ok {
+		return nil, 0, types.ErrItemNotFound("Unable to find raw data report with request ID (%d) external ID (%d) from (%s)", int64(env.requestID), externalDataID, validatorAddress.String())
```
taobun: Remove int64(...)

chain/x/zoracle/exec_env_test.go
```diff
@@ -182,8 +182,8 @@ func TestRequestExternalData(t *testing.T) {
 
 	env, err := NewExecutionEnvironment(ctx, keeper, 1)
 	require.Nil(t, err)
-	envErr := env.RequestExternalData(1, 42, []byte("prepare32"))
-	require.Nil(t, envErr)
+	env.RequestExternalData(1, 42, []byte("prepare32"))
```
taobun: RequestExternal can fail

taobun (changes requested):

taobun (approved):

sorawit (approved):

PR #981, opened 2020-03-24
Implement caching layer for block information

### LINKED ISSUE #914: Implement caching layer for block information
Add Block table to the database and make sure it gets updated at every block. The table should contain:

- Block height (primary key)
- Transaction count
- Timestamp
- Proposer (foreign key to validator table)
- BlockHash

prin-r:
Fix #914

### DIFF
--- a/chain/app/db_app.go
+++ b/chain/app/db_app.go
@@ -1,6 +1,7 @@
 package app
 
 import (
+	"fmt"
 	"io"
 	"time"
 
@@ -168,6 +169,13 @@ func (app *dbBandApp) BeginBlock(req abci.RequestBeginBlock) (res abci.ResponseB
 	if err != nil {
 		panic(err)
 	}
+
+	app.dbBand.AddBlock(
+		req.Header.GetHeight(),
+		time.Now(),
+		fmt.Sprintf("%X", req.Header.GetProposerAddress()),
+		req.GetHash(),
+	)
 
 	return res
 }
--- a/chain/app/db_app.go
+++ b/chain/app/db_app.go
@@ -1,7 +1,6 @@
 package app
 
 import (
-	"fmt"
 	"io"
 	"time"
 
@@ -173,7 +172,7 @@ func (app *dbBandApp) BeginBlock(req abci.RequestBeginBlock) (res abci.ResponseB
 	app.dbBand.AddBlock(
 		req.Header.GetHeight(),
 		time.Now(),
-		fmt.Sprintf("%X", req.Header.GetProposerAddress()),
+		req.Header.GetProposerAddress(),
 		req.GetHash(),
 	)
 
--- a/chain/app/db_app.go
+++ b/chain/app/db_app.go
@@ -171,7 +171,7 @@ func (app *dbBandApp) BeginBlock(req abci.RequestBeginBlock) (res abci.ResponseB
 
 	app.dbBand.AddBlock(
 		req.Header.GetHeight(),
-		app.DeliverContext.BlockHeader().Time,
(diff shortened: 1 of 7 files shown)


chain/db/table.go
```diff
@@ -46,3 +46,10 @@ type DataSourceRevision struct {
 	BlockHeight    int64  `gorm:"not null"`
 	TxHash         []byte `gorm:"not null"`
 }
+
+type Block struct {
+	Height    int64  `gorm:"primary_key;auto_increment:false"`
+	Timestamp int64  `gorm:"not null"`
+	Proposer  string `gorm:"not null"`
+	BlockHash string `gorm:"unique;not null"`
```
sorawit: I don't think we need to say it's unique?

chain/db/block.go
```diff
@@ -0,0 +1,21 @@
...
+
+func (b *BandDB) AddBlock(
+	height int64,
+	timestamp time.Time,
+	proposer string,
+	blockHash common.HexBytes,
+) error {
+	return b.tx.Create(&Block{
+		Height:    height,
+		Timestamp: timestamp.Unix(),
```
sorawit: Let's do more precise timestamp here.

sorawit (changes requested):

chain/db/table.go
taobun: We can store block hash as []byte

chain/db/block.go
taobun: What does it mean?

chain/app/db_app.go
```diff
@@ -157,6 +158,13 @@ func (app *dbBandApp) BeginBlock(req abci.RequestBeginBlock) (res abci.ResponseB
 		panic(err)
 	}
 
+	app.dbBand.AddBlock(
+		req.Header.GetHeight(),
+		time.Now(),
+		fmt.Sprintf("%X", req.Header.GetProposerAddress()),
```
taobun: Wow please change my code in validator.go too

chain/db/block.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"time"
+
+	"github.com/tendermint/tendermint/libs/common"
+)
+
+func (b *BandDB) AddBlock(
+	height int64,
+	timestamp time.Time,
+	proposer string,
```
taobun: Receive proposer as cmn.HexBytes and cast to string in function instead.

taobun (changes requested):

chain/app/db_app.go
```diff
@@ -157,6 +157,13 @@ func (app *dbBandApp) BeginBlock(req abci.RequestBeginBlock) (res abci.ResponseB
 		panic(err)
 	}
 
+	app.dbBand.AddBlock(
+		req.Header.GetHeight(),
+		time.Now(),
```
taobun: Use block time from context

taobun (changes requested):

chain/app/db_app.go
```diff
@@ -157,6 +157,13 @@ func (app *dbBandApp) BeginBlock(req abci.RequestBeginBlock) (res abci.ResponseB
 		panic(err)
 	}
 
+	app.dbBand.AddBlock(
+		req.Header.GetHeight(),
+		app.DeliverContext.BlockHeader().Time,
```
taobun: `app.DeliverContext.BlockTime()`

taobun (changes requested):

taobun (approved):

sorawit (approved):

PR #983, opened 2020-03-25
Implement caching layer for validator table

### LINKED ISSUE #943: Implement caching layer for validator table
Extend what @taobun has written to make it fully featured, with moniker, description, etc. Make sure it's updated whenever there's a message to create validator or edit validator.

prin-r:
Fix #943


chain/app/db_app.go
```diff
@@ -71,6 +71,24 @@ func (app *dbBandApp) InitChain(req abci.RequestInitChain) abci.ResponseInitChai
...
 					createMsg.ValidatorAddress,
 					createMsg.PubKey,
+					createMsg.Description.Moniker,
+					createMsg.Description.Identity,
+					createMsg.Description.Website,
+					createMsg.Description.Details,
+					createMsg.Commission.Rate.String(),
+					createMsg.Commission.MaxRate.String(),
+					createMsg.Commission.MaxChangeRate.String(),
+					sdk.NewCoin("uband", createMsg.MinSelfDelegation).String(),
```
sorawit: Let's not hard-code `uband`. We can just store it add is. The frontend will decide if it wants to add the unit.

chain/db/Delegation.go
```diff
@@ -0,0 +1,17 @@
+package db
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func (b *BandDB) AddDelegation(
```
sorawit, line 8: Please also handle `MsgDelegate` to make sure we keep track of delegations correctly.

sorawit (changes requested):

chain/db/Delegation.go
```diff
@@ -0,0 +1,34 @@
+package db
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/staking"
+)
+
+func (b *BandDB) AddDelegation(
```
sorawit, line 8: This is still not correct. You will need to handle slashing, inflation, etc to make delegate works. Let's comment out all the code for now.

sorawit (changes requested):

chain/db/validator.go
```diff
@@ -88,3 +109,57 @@ func (b *BandDB) ClearOldVotes(currentHeight int64) error {
 	}
 	return nil
 }
+
+func (b *BandDB) GetValidator(validator Validator) (Validator, error) {
```
taobun: Change signature to GetValidator(validator sdk,ValAddress) (Validator, bool)

chain/db/validator.go
```diff
@@ -88,3 +109,57 @@ func (b *BandDB) ClearOldVotes(currentHeight int64) error {
...
+	err := b.tx.First(&validator).Error
```
taobun: return true if found and false if error/not found

taobun (changes requested):

chain/db/validator.go
```diff
@@ -88,3 +109,63 @@ func (b *BandDB) ClearOldVotes(currentHeight int64) error {
 	}
 	return nil
 }
+
+func (b *BandDB) GetValidator(validator sdk.ValAddress) (Validator, bool) {
+	validatorSturct := Validator{OperatorAddress: validator.String()}
```
taobun: validatorStruct

taobun (changes requested):

chain/db/validator.go
```diff
@@ -88,3 +109,63 @@ func (b *BandDB) ClearOldVotes(currentHeight int64) error {
...
+	err := b.tx.First(&validatorSturct).Error
+	if err == nil {
```
taobun: return validatorStruct, err == nil

taobun (changes requested):

chain/db/validator.go
```diff
@@ -88,3 +109,58 @@ func (b *BandDB) ClearOldVotes(currentHeight int64) error {
...
+	return validatorStruct, err == nil
+}
+
+func (b *BandDB) handleMsgEditValidator(msg staking.MsgEditValidator) error {
+	validator, isFound := b.GetValidator(msg.ValidatorAddress)
+	if !isFound {
+		return fmt.Errorf(fmt.Sprintf("validator %s has not exist.", msg.ValidatorAddress.String()))
+	}
+
+	if msg.Description.Moniker != "[do-not-modify]" {
```
sorawit: Why do we type out string here? Let's use `DoNotModifyDesc` variable?

taobun (dismissed):

chain/db/table.go
```diff
@@ -14,12 +14,27 @@ type Event struct {
 	Name string
 }
 
-type Validator struct {
+type Delegation struct {
```
sorawit: Let's comment out the whole table.

sorawit (changes requested):

chain/db/validator.go
taobun: https://github.com/cosmos/cosmos-sdk/blob/release/v0.37.8/x/staking/alias.go

taobun (dismissed):

chain/db/validator.go
```diff
@@ -88,3 +109,58 @@ func (b *BandDB) ClearOldVotes(currentHeight int64) error {
...
+func (b *BandDB) GetValidator(validator sdk.ValAddress) (Validator, bool) {
+	validatorStruct := Validator{OperatorAddress: validator.String()}
+	err := b.tx.First(&validatorStruct).Error
+	return validatorStruct, err == nil
+}
+
+func (b *BandDB) handleMsgEditValidator(msg staking.MsgEditValidator) error {
+	validator, isFound := b.GetValidator(msg.ValidatorAddress)
+	if !isFound {
+		return fmt.Errorf(fmt.Sprintf("validator %s has not exist.", msg.ValidatorAddress.String()))
```
sorawit: does not exist

chain/db/validator.go
```diff
@@ -88,3 +109,58 @@ func (b *BandDB) ClearOldVotes(currentHeight int64) error {
...
+		validator.MinSelfDelegation = msg.MinSelfDelegation.ToDec().String()
+	}
+
+	return b.tx.Save(&validator).Error
+}
+
+func (b *BandDB) handleMsgCreateValidator(msg staking.MsgCreateValidator) error {
+	_, isFound := b.GetValidator(msg.ValidatorAddress)
+	if isFound {
+		return fmt.Errorf(fmt.Sprintf("validator %s has already exist.", msg.ValidatorAddress.String()))
```
sorawit: Why do we need to do this? I don't think it's even possible to not found.

sorawit (changes requested):

chain/db/validator.go
taobun: to found ?

sorawit (approved):

taobun (approved):

PR #984, opened 2020-03-25
Caching layer: oracle scripts and revisions

### LINKED ISSUE #917: Implement caching layer for oracle script info
The oracle script table should contain
- ID (pk)
- name
- description
- owner
- request fee
- last update

Also add oracle script revision table that should contain
- ID (pk)
- Revision number (pk)
- Name
- timestamp
- block height
- tx hash

evilpeach:
Fixes #917

### DIFF
--- a/chain/db/oracle_script.go
+++ b/chain/db/oracle_script.go
@@ -22,7 +22,7 @@ func createOracleScript(
 		Description: description,
 		Owner:       owner.String(),
 		Code:        code,
-		LastUpdated: blockTime.Unix(),
+		LastUpdated: blockTime.UnixNano() / int64(time.Millisecond),
 	}
 }
 
@@ -51,7 +51,7 @@ func (b *BandDB) AddOracleScript(
 	return b.tx.Create(&OracleScriptRevision{
 		OracleScriptID: id,
 		Name:           name,
-		Timestamp:      blockTime.Unix(),
+		Timestamp:      blockTime.UnixNano() / int64(time.Millisecond),
 		BlockHeight:    blockHeight,
 		TxHash:         txHash,
 	}).Error
@@ -94,7 +94,7 @@ func (b *BandDB) handleMsgEditOracleScript(
 	return b.tx.Create(&OracleScriptRevision{
 		OracleScriptID: int64(msg.OracleScriptID),
 		Name:           msg.Name,
-		Timestamp:      b.ctx.BlockTime().Unix(),
+		Timestamp:      b.ctx.BlockTime().UnixNano() / int64(time.Millisecond),
 		BlockHeight:    b.ctx.BlockHeight(),
 		TxHash:         txHash,
 	}).Error
--- a/chain/db/oracle_script.go
+++ b/chain/db/oracle_script.go
@@ -1,0 +1,101 @@
+package db
+
+import (
+	"errors"
+	"strconv"
+	"time"
+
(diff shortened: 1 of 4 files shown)


chain/db/oracle_script.go
```diff
@@ -0,0 +1,101 @@
...
+	code []byte,
+	blockTime time.Time,
+) OracleScript {
+	return OracleScript{
+		ID:          id,
+		Name:        name,
+		Description: description,
+		Owner:       owner.String(),
+		Code:        code,
+		LastUpdated: blockTime.Unix(),
```
sorawit: Please use millisecond.

chain/db/oracle_script.go
```diff
@@ -0,0 +1,101 @@
...
+	)
+	err := b.tx.Create(&oracleScript).Error
+	if err != nil {
+		return err
+	}
+
+	return b.tx.Create(&OracleScriptRevision{
+		OracleScriptID: id,
+		Name:           name,
+		Timestamp:      blockTime.Unix(),
```
sorawit: ditto

sorawit (changes requested):
Please fix merge conflict.

chain/db/table.go
```diff
@@ -47,6 +47,24 @@ type DataSourceRevision struct {
...
+	LastUpdated int64  `gorm:"not null"`
+}
+
+type OracleScriptRevision struct {
+	OracleScriptID int64  `gorm:"primary_key;auto_increment:false"`
+	RevisionNumber int64  `gorm:"primary_key"`
+	Name           string `gorm:"not null"`
+	Timestamp      int64  `gorm:"not null"`
+	BlockHeight    int64  `gorm:"not null"`
+	TxHash         []byte `gorm:"not null"`
```
sorawit: Add fk to transaction

sorawit (changes requested):

chain/db/table.go
sorawit: @taobun should we add fk here? It will not work with oracle scripts generated during genesis......

taobun: If we allow to have txHash 000000... as genesis state, it sounds make sense?

sorawit: Yeah let's do that.

sorawit: Three choices:
- Don't do FK
- Use dummy TX like you suggest
- Make `txHash` column nullable. NULL does not need to match FK.

taobun: The answer is Make txHash column nullable. NULL does not need to match FK.

taobun: watch #995 as example

taobun (approved):
LGTM

chain/db/table.go
```diff
@@ -47,6 +47,24 @@ type DataSourceRevision struct {
...
+	Owner       string `gorm:"not null"`
+	Code        []byte `gorm:"not null"`
+	LastUpdated int64  `gorm:"not null"`
+}
+
+type OracleScriptRevision struct {
+	OracleScriptID int64  `gorm:"primary_key;auto_increment:false"`
+	RevisionNumber int64  `gorm:"primary_key"`
+	Name           string `gorm:"not null"`
+	Timestamp      int64  `gorm:"not null"`
```
sorawit, line 63: We will need to remove this, but let's do it on a separate diff.

sorawit (approved):

PR #987, opened 2020-03-25
Design and implement borsh decode function

### LINKED ISSUE #932: Design and implement borsh decode function
The function should take three inputs: `schema: string`, `cls: string`, and `data: Buffer.t` and return `option(list((string, string)))`. The first element is field name and the second element is the value.  Return None if the decode fails. Nested decode should be flattened.

taobun:
fixed #932


scan/src/bindings/Borsh.re
```diff
@@ -34,3 +34,48 @@ function() {
 }
 |}
 ];
+
+// Test by this test
```
sorawit: Instead of having this as comment, let's create an actual test file.

sorawit (changes requested):

sorawit (approved):

scan/src/bindings/Borsh.re
```diff
@@ -34,3 +34,42 @@ function() {
...
+        window[className] = gen(t.fields.map(x => x[0]))
+        t.fields = t.fields.map(x => {
+          x[1] = x[1].replace(/^\w/, c => c.toUpperCase());
+          return x
+        })
+        schemaMap.set(window[className], t)
+      }
+    }
+    let model = window[cls]
+    let new_value = borsh.deserialize(schemaMap, model, data)
```
sorawit: camelcase?

sorawit (changes requested):

sorawit (approved):

evilpeach (approved):
LGTM

PR #995, opened 2020-03-26
Update data source revistion table to have null tx

taobun:


chain/db/table.go
```diff
@@ -44,7 +44,7 @@ type DataSourceRevision struct {
 	Name           string `gorm:"not null"`
 	Timestamp      int64  `gorm:"not null"`
 	BlockHeight    int64  `gorm:"not null"`
-	TxHash         []byte `gorm:"not null"`
+	TxHash         []byte `sql:"default:null"`
```
sorawit, line 47: I don't think we need to put in default?

sorawit (changes requested):

chain/db/table.go
sorawit: NULL should be the automatic default for table that is nullable, no?

taobun: If not set this tag, I cannot add nil of []byte as null in table.

sorawit (dismissed):

evilpeach (dismissed):

taobun:
Something wrong when fix conflict new pr #997

PR #1001, opened 2020-03-26
Example decoder of result

taobun:


sorawit (changes requested):
Please add some tests

sorawit (approved):
Looks great

PR #1059, opened 2020-03-29
chain: Perform raw report duplication check in ValidateBasic

sorawit:
The check can be performed without needing access to state, so let's check in ValidateBasic. I also don't think the requirement for increasing order matters much here. While it does help with performance, it's extremely minimal and non-significant.


chain/x/zoracle/internal/types/msgs.go
```diff
@@ -135,6 +135,13 @@ func (msg MsgReportData) ValidateBasic() sdk.Error {
 	if msg.DataSet == nil || len(msg.DataSet) == 0 {
 		return ErrInvalidBasicMsg("MsgReportData: Data set must not be empty.")
 	}
+	for idx1, rawReport1 := range msg.DataSet {
+		for _, rawReport2 := range msg.DataSet[idx1+1:] {
+			if rawReport1.ExternalDataID == rawReport2.ExternalDataID {
+				return ErrInvalidBasicMsg("MsgReportData: External IDs in dataset must be unique.")
+			}
+		}
+	}
```
Benzbeeb, line 144: Can we check unique by use map? I think that way is easy to read and better in time complexity

Benzbeeb (changes requested):

taobun (approved):

Benzbeeb (approved):

PR #1061, opened 2020-03-29
chain: Convert params to uint64 and introduce one function to read them

sorawit:
All parameters are in the form of Max... or Gas..., so it makes sense to make all of them `uint64`. It also makes sense to just have one function to get/set, rather have the number of functions scales according to number of params.

### DIFF
--- a/chain/x/zoracle/internal/keeper/request.go
+++ b/chain/x/zoracle/internal/keeper/request.go
@@ -33,10 +33,10 @@ func (k Keeper) AddRequest(
 		return 0, types.ErrItemNotFound("AddRequest: Unknown oracle script ID %d.", oracleScriptID)
 	}
 
-	if int64(len(calldata)) > k.MaxCalldataSize(ctx) {
+	if uint64(len(calldata)) > k.GetParam(ctx, types.KeyMaxCalldataSize) {
 		return 0, types.ErrBadDataValue(
 			"AddRequest: Calldata size (%d) exceeds the maximum size (%d).",
-			len(calldata), int(k.MaxCalldataSize(ctx)),
+			len(calldata), int(k.GetParam(ctx, types.KeyMaxCalldataSize)),
 		)
 	}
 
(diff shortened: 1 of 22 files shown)


Benzbeeb (dismissed):

chain/x/zoracle/internal/keeper/result.go
```diff
@@ -10,10 +10,10 @@ func (k Keeper) AddResult(
 	ctx sdk.Context, requestID types.RequestID, oracleScriptID types.OracleScriptID,
 	calldata []byte, result []byte,
 ) sdk.Error {
-	if int64(len(result)) > k.MaxResultSize(ctx) {
+	if uint64(len(result)) > k.GetParam(ctx, types.KeyMaxResultSize) {
 		return types.ErrBadDataValue(
 			"AddResult: Result size (%d) exceeds the maximum size (%d).",
-			len(result), int(k.MaxResultSize(ctx)),
+			len(result), int(k.GetParam(ctx, types.KeyMaxResultSize)),
```
taobun: In fmt we need to cast uint64 to int ?

chain/x/zoracle/internal/keeper/request.go
```diff
@@ -33,10 +33,10 @@ func (k Keeper) AddRequest(
 		return 0, types.ErrItemNotFound("AddRequest: Unknown oracle script ID %d.", oracleScriptID)
 	}
 
-	if int64(len(calldata)) > k.MaxCalldataSize(ctx) {
+	if uint64(len(calldata)) > k.GetParam(ctx, types.KeyMaxCalldataSize) {
 		return 0, types.ErrBadDataValue(
 			"AddRequest: Calldata size (%d) exceeds the maximum size (%d).",
-			len(calldata), int(k.MaxCalldataSize(ctx)),
+			len(calldata), int(k.GetParam(ctx, types.KeyMaxCalldataSize)),
```
taobun: ditto

taobun (changes requested):

chain/x/zoracle/internal/keeper/result.go
sorawit: Good catch

taobun (approved):

Benzbeeb (approved):

PR #1073, opened 2020-03-30
Account balance table in caching layer

### LINKED ISSUE #1024: Account balance table in caching layer. Research + Implement.
Research and define how to track balance of users in our database. Consult plan with @taobun to make sure it works in all cases. Some ideas:
- Track transfer event
- Check if messages or transactions involve what users and reload their balance

Things to be careful of
- Genesis account
- Transaction fees
- etc....

taobun:
fixed #1024

### DIFF
--- a/chain/db/account.go
+++ b/chain/db/account.go
@@ -4,13 +4,13 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-func (b *BandDB) AddOrCreateAccount(address sdk.AccAddress, balance sdk.Coins) error {
+func (b *BandDB) SetAccountBalance(address sdk.AccAddress, balance sdk.Coins) error {
 	return b.tx.Where(Account{Address: address.String()}).
 		Assign(Account{Balance: balance.String()}).
 		FirstOrCreate(&Account{}).Error
 }
 
-func (b *BandDB) DecreaseBalance(address sdk.AccAddress, balance sdk.Coins) error {
+func (b *BandDB) DecreaseAccountBalance(address sdk.AccAddress, balance sdk.Coins) error {
 	account := Account{Address: address.String()}
 	err := b.tx.First(&account).Error
 	if err != nil {
--- a/chain/db/account.go
+++ b/chain/db/account.go
@@ -4,13 +4,17 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-func (b *BandDB) SetAccountBalance(address sdk.AccAddress, balance sdk.Coins) error {
+func (b *BandDB) SetAccountBalance(
+	address sdk.AccAddress, balance sdk.Coins, blockHeight int64,
+) error {
 	return b.tx.Where(Account{Address: address.String()}).
-		Assign(Account{Balance: balance.String()}).
+		Assign(Account{Balance: balance.String(), UpdatedHeight: blockHeight}).
 		FirstOrCreate(&Account{}).Error
 }
 
-func (b *BandDB) DecreaseAccountBalance(address sdk.AccAddress, balance sdk.Coins) error {
+func (b *BandDB) DecreaseAccountBalance(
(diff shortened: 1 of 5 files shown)


chain/db/account.go
```diff
@@ -0,0 +1,25 @@
+package db
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func (b *BandDB) AddOrCreateAccount(address sdk.AccAddress, balance sdk.Coins) error {
```
sorawit: Should we calll this `CreateOrUpdateAccount`? Or `SetAccountBalance`?

chain/db/db.go
```diff
@@ -33,21 +35,42 @@ func NewDB(dialect, path string, metadata map[string]string) (*BandDB, error) {
...
+	)
+
+	db.Model(&Transaction{}).AddForeignKey(
+		"block_height",
+		"blocks(height)",
+		"RESTRICT",
+		"RESTRICT",
+	)
+
+	db.Model(&Transaction{}).AddForeignKey(
```
sorawit, line 72: Is it possible that we add transaction before account and violate fk constraint?

chain/db/db.go
```diff
@@ -349,3 +358,45 @@ func (b *BandDB) HandleMessage(txHash []byte, msg sdk.Msg, events map[string]str
 
 	return jsonMap, nil
 }
+
+func (b *BandDB) GetInvolvedAccounts(tx auth.StdTx) []sdk.AccAddress {
+	involvedAccounts := make([]sdk.AccAddress, 0)
+	for _, msg := range tx.GetMsgs() {
+		switch msg := msg.(type) {
+		case zoracle.MsgRequestData:
+			involvedAccounts = append(involvedAccounts, msg.Sender)
```
sorawit, line 389: All data source owners will get fees, right?

sorawit (changes requested):

chain/db/db.go
```diff
@@ -349,3 +358,45 @@ func (b *BandDB) HandleMessage(txHash []byte, msg sdk.Msg, events map[string]str
...
+			involvedAccounts = append(involvedAccounts, msg.DelegatorAddress)
+		case dist.MsgWithdrawDelegatorReward:
+			involvedAccounts = append(involvedAccounts, msg.DelegatorAddress)
+		case dist.MsgWithdrawValidatorCommission:
+			involvedAccounts = append(involvedAccounts, sdk.AccAddress(msg.ValidatorAddress))
+		case gov.MsgDeposit:
+			involvedAccounts = append(involvedAccounts, msg.Depositor)
+		case gov.MsgSubmitProposal:
+			involvedAccounts = append(involvedAccounts, msg.Proposer)
+		default:
```
sorawit, line 427: Now that we handle a lot of messages. Let's make sure we have case for all messages, and panic for unknown messages, so we fail fast.

sorawit (changes requested):

chain/db/table.go
```diff
@@ -2,18 +2,35 @@ package db
...
+	GasFee      string          `gorm:"not null"`
+	Sender      string          `gorm:"not null"`
+	Success     bool            `gorm:"not null"`
+	BlockHeight int64           `gorm:"not null"`
+	Messages    json.RawMessage `sql:"json;not null"`
+}
+
+type Account struct {
+	Address string `gorm:"primary_key"`
+	Balance string `gorm:"not null"`
```
sorawit: Let's store UpdatedHeight here. I think it will help us debug a lot in the future.

sorawit (changes requested):

chain/db/db.go
taobun: A sender is needed to declare before in chain (Cannot send a transaction without account number and sequence number)

sorawit (dismissed):

Benzbeeb (dismissed):

sorawit (approved):
Let's ship it

Benzbeeb (approved):

Issue #1035, opened 2020-03-28
Figure out our plan to deal with 1317 CORS issue


sorawit:
Do we need to add a flag when running REST server? Is it an issue with Varnish? Do we need to run another proxy? How to fix that. Present what you experiment and solution

PR #1075, opened 2020-03-30
scan: Account context

### LINKED ISSUE #1037: Implement context to keep user private key
We will have barebone login features. This should act like a global context that people can dispatch the following actions.
- Connect(mnemonic) -> login with the given key
- Disconnect -> logout

The context should expose state as the user address to other components.

evilpeach:
Fixes #1037


scan/src/contexts/AccountContext.re
```diff
@@ -0,0 +1,36 @@
...
+};
+
+type a =
+  | Connect(string)
+  | Disconnect;
+
+let reducer = _ =>
+  fun
+  | Connect(mnemonic) => {
+      let bandchain = BandWeb3.network("http://localhost:8010", "bandchain");
```
sorawit: Let's do this globally up top.

sorawit (changes requested):

scan/src/contexts/AccountContext.re
sorawit: This http://localhost:8010 will need to be revisited #1035

evilpeach: Sure

sorawit (approved):

prin-r (approved):

PR #1089, opened 2020-04-01
scan: user account ui and wire up with account context

### LINKED ISSUE #1040: Show user account and implement the way for connect
The way can be very simple, like using prompt. Consult @evilpeach or @sorawit for how to do interop if necessary. By the end of this ticket user should be able to connect by entering mnemonic. The prompt should also suggest some random mnemonic.

evilpeach:
Fixes #1040


scan/src/components/UserAccount.re
```diff
@@ -0,0 +1,152 @@
...
+      <HSpacing size=Spacing.sm />
+      <img src=Images.disconnectIcon className=Styles.logo />
+    </div>;
+  };
+};
+
+module FaucetBtn = {
+  [@react.component]
+  let make = () => {
+    <div className=Styles.faucetBtn>
```
sorawit: Let's alert or console.log after in onClick for now.

sorawit (changes requested):

scan/src/components/UserAccount.re
evilpeach: Sure

sorawit (approved):
Looks great

prin-r (approved):

PR #1115, opened 2020-04-03
Remove prepare gas from MsgRequestData 

### LINKED ISSUE #1018: Remove prepare gas from MsgRequestData
Instead of needing to pass prepare gas explicitly. We should just use the current remaining gas as the gas to run. Then consume whatever amount is used later. Also update necessary test cases.

Benzbeeb:
Fixed #1018

### DIFF
--- a/chain/x/zoracle/client/cli/tx.go
+++ b/chain/x/zoracle/client/cli/tx.go
@@ -61,7 +61,7 @@ func GetCmdRequest(cdc *codec.Codec) *cobra.Command {
 		Long: strings.TrimSpace(
 			fmt.Sprintf(`Make a new request via an existing oracle script with the configuration flags.
 Example:
-$ %s tx zoracle request 1 -c 1234abcdef -r 4 -v 3 -x 20 -w 50 -g 5000 --from mykey
+$ %s tx zoracle request 1 -c 1234abcdef -r 4 -v 3 -x 20 -g 5000 --from mykey
 $ %s tx zoracle request 1 --calldata 1234abcdef --requested-validator-count 4 --sufficient-validator-count 3 --expiration 20 --execute-gas 5000 --from mykey
 `,
 				version.ClientName, version.ClientName,
--- a/chain/x/zoracle/client/cli/tx.go
+++ b/chain/x/zoracle/client/cli/tx.go
@@ -62,7 +62,7 @@ func GetCmdRequest(cdc *codec.Codec) *cobra.Command {
 			fmt.Sprintf(`Make a new request via an existing oracle script with the configuration flags.
 Example:
 $ %s tx zoracle request 1 -c 1234abcdef -r 4 -v 3 -x 20 -w 50 -g 5000 --from mykey
-$ %s tx zoracle request 1 --calldata 1234abcdef --requested-validator-count 4 --sufficient-validator-count 3 --expiration 20 --prepare-gas 50 --execute-gas 5000 --from mykey
+$ %s tx zoracle request 1 --calldata 1234abcdef --requested-validator-count 4 --sufficient-validator-count 3 --expiration 20 --execute-gas 5000 --from mykey
 `,
 				version.ClientName, version.ClientName,
 			),
--- a/chain/x/zoracle/client/cli/tx.go
+++ b/chain/x/zoracle/client/cli/tx.go
@@ -28,7 +28,6 @@ const (
 	flagRequestedValidatorCount  = "requested-validator-count"
 	flagSufficientValidatorCount = "sufficient-validator-count"
 	flagExpiration               = "expiration"
-	flagPrepareGas               = "prepare-gas"
 	flagExecuteGas               = "execute-gas"
 )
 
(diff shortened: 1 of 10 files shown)


chain/x/zoracle/handler_test.go
```diff
@@ -276,41 +274,12 @@ func TestRequestInvalidDataSource(t *testing.T) {
 	require.False(t, got.IsOK())
 }
 
-func TestRequestWithPrepareGasExceed(t *testing.T) {
```
sorawit: Do not remove this. Modify it to test what we want.

sorawit (changes requested):

chain/x/zoracle/handler_test.go
```diff
@@ -299,18 +299,17 @@ func TestRequestWithPrepareGasExceed(t *testing.T) {
 	dataSource := keep.GetTestDataSource()
 	keeper.SetDataSource(ctx, 1, dataSource)
 
-	// set prepare gas to 3 (not enough for using) then it occurs error.
-	msg := types.NewMsgRequestData(1, calldata, 2, 2, 100, 3, 1000000, sender)
+	msg := types.NewMsgRequestData(1, calldata, 2, 2, 100, 1000000, sender)
 
-	got := handleMsgRequestData(ctx, keeper, msg)
-	require.False(t, got.IsOK())
+	require.Panics(t, func() { handleMsgRequestData(ctx, keeper, msg) }, "The code did not panic")
```
sorawit, line 310: Make the failing comment clear. Why do you expect a panic?

chain/x/zoracle/handler_test.go
```diff
@@ -277,11 +275,13 @@ func TestRequestInvalidDataSource(t *testing.T) {
 }
 
 func TestRequestWithPrepareGasExceed(t *testing.T) {
-	// Setup test environment
+
 	ctx, keeper := keep.CreateTestInput(t, false)
 
 	ctx = ctx.WithBlockHeight(2)
 	ctx = ctx.WithBlockTime(time.Unix(int64(1581589790), 0))
+	// gas consumed before prepare = 1732482
```
sorawit: What does this mean? Gas consumed before prepare??

I think you should make clear how much gas the code's prepare function is expected to consume. Then test with both with enough and not-enough gas.

chain/x/zoracle/handler_test.go
```diff
@@ -277,11 +275,13 @@ func TestRequestInvalidDataSource(t *testing.T) {
 }
 
 func TestRequestWithPrepareGasExceed(t *testing.T) {
-	// Setup test environment
```
sorawit: Don't remove code without a good reason.

sorawit (changes requested):

chain/x/zoracle/client/cli/tx.go
```diff
@@ -58,7 +57,7 @@ func GetTxCmd(storeKey string, cdc *codec.Codec) *cobra.Command {
 // GetCmdRequest implements the request command handler.
 func GetCmdRequest(cdc *codec.Codec) *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "request [oracle-script-id] (-c [calldata]) (-r [requested-validator-count]) (-v [sufficient-validator-count]) (-x [expiration]) (-w [prepare-gas]) (-g [execute-gas])",
+		Use:   "request [oracle-script-id] (-c [calldata]) (-r [requested-validator-count]) (-v [sufficient-validator-count]) (-x [expiration]) (-g [execute-gas])",
 		Short: "Make a new data request via an existing oracle script",
 		Args:  cobra.ExactArgs(1),
 		Long: strings.TrimSpace(
```
sorawit, line 61: Change example below too....

sorawit (changes requested):

chain/x/zoracle/handler_test.go
```diff
@@ -299,10 +297,18 @@ func TestRequestWithPrepareGasExceed(t *testing.T) {
...
 
+	gasConsumedBeforeHandle := ctx.GasMeter().GasConsumed()
 	got := handleMsgRequestData(ctx, keeper, msg)
+	gasConsumedAfterHandle := ctx.GasMeter().GasConsumed()
+
+	// the code's prepare function is expected to consume 103 gas
+	ctx = ctx.WithGasMeter(sdk.NewGasMeter(gasConsumedAfterHandle - gasConsumedBeforeHandle - 103))
+
+	// handleMsgRequestData will panic when the gas limit doesn't enough to consume
+	require.Panics(t, func() { handleMsgRequestData(ctx, keeper, msg) }, "The code did not panic")
```
taobun, line 310: In my opinion this code panic because the gas doesn't enough to consume for execution not run prepare owasm code. 

*Note, I think if gas prepare is not enough it will return error not panic code.

taobun (changes requested):

sorawit:
This turns out to be quite complicated. Let's close this for now. #1225 will take this forward.

PR #1118, opened 2020-04-03
scan: Remove age (... ago) in all places

tansawit:
Fixes: #1041

PR #1125, opened 2020-04-03
Wireup top cards latest block and active validators

### LINKED ISSUE #994: Wireup top cards latest block and active validators
- Latest block should come from latest block hook. Should also update the block proposer.
- Active validators card should be wired up.
- Also remove "minutes ago" on block card.

Benzbeeb:
Fixed #994


Benzbeeb:
PR #1118 already fix this

PR #1129, opened 2020-04-04
Validator stake & delegation table in caching layer. 

### LINKED ISSUE #1026: Validator stake & delegation table in caching layer. Research + Implement.
Research design and implement the way to track delegation share to total share in cacher. Must fully understand lifecycle of validator delegation and staking. Expect the task owner to talk through that clearly and implement respective cacher.

### LINKED ISSUE #1119: Update account balance during EndBlock for mature undelegation
NOTE: There are no transfer events. The accounts are set directly! See code in `cosmos-sdk/x/staking/handler.go`'s `EndBlocker` function. Mature unbonding tokens will be sent to users. Please verify that your code works correctly by delegating and unbonding.

taobun:
fixed #1026 
fixed #1119

- [x] Create validator & delegate in genesis
- [x] Handle create validator msg
- [x] Handle delegate msg
- [x] Handle redelegate msg
- [x] Handle undelegate(unbond) msg
- [x] Update account balance after unbonding period correctly


chain/db/endblock.go
```diff
@@ -0,0 +1,70 @@
...
+					panic(err)
+				}
+				rawResult = result.Data
+			}
+			err = b.ResolveRequest(requestID, resolveStatus, rawResult)
+			if err != nil {
+				panic(err)
+			}
+		}
+	case staking.EventTypeCompleteUnbonding:
```
sorawit, line 53: So I think this fixes #1119?

chain/db/delegation.go
```diff
@@ -0,0 +1,96 @@
...
+	delegatorAddress sdk.AccAddress,
+	validatorAddress sdk.ValAddress,
+) error {
+	return b.tx.Delete(Delegation{
+		DelegatorAddress: delegatorAddress.String(),
+		ValidatorAddress: validatorAddress.String(),
+	}).Error
+}
+
+func (b *BandDB) delegate(
```
sorawit, line 32: Can a validator delegate for himself? Does `SelfDelegation` need to increase?

chain/db/slashing.go
```diff
@@ -0,0 +1,11 @@
+package db
```
sorawit, line 1: Let's handle the case when a validator got jailed too. Jail should become true in that case?

sorawit (changes requested):
How do we compute validator stake? I the PR say "Validator stake", but its't this PR only for delegation share?

chain/db/delegation.go
taobun: SelfDelegation is just a delegation record, not a variable in validator struct, logic query delegation by validator to compare with MinSelfDelegation

chain/db/endblock.go
taobun: yep

chain/db/delegation.go
sorawit: Make sense.

chain/app/db_app.go
```diff
@@ -244,45 +250,27 @@ func (app *dbBandApp) EndBlock(req abci.RequestEndBlock) (res abci.ResponseEndBl
...
-			err := app.dbBand.ResolveRequest(requestID, resolveStatus, rawResult)
-			if err != nil {
-				panic(err)
-			}
-		}
+		app.dbBand.HandleEndblockEvent(event)
 	}
 
+	// Update bonding validator status
+	app.StakingKeeper.IterateValidators(app.DeliverContext,
```
sorawit: If we are looping like this and already fetch all the data, then why don't we use this loop to update other info as well? Tokens, Jails, DelegatorShares, etc can be updated here. and we will be very sure that we won't mess up. Would that be better?

sorawit (changes requested):

chain/app/db_app.go
sorawit: Now that I think about it. Let's just remove this status thing. It looks quite inefficient and I don't think we need to show it for now. We can certainly revisit later.

sorawit (changes requested):

sorawit (dismissed):

sorawit:
@taobun Please resolve conflict due to `Pubkey` becoming `string`.

sorawit (approved):

tansawit (approved):

PR #1130, opened 2020-04-04
chain: Patch to v0.38 with IBC

sorawit:
Now that phase 0 is release. It's time to go into the more fun territory.

Things already tested
- Send transactions with bandoracled reporting
- Work with docker-compose
- Work with db (blocks / transactions)

Things not done yet.
- Patch db to support new messages
- Integrate IBC
- Bring back proof gen


chain/Makefile
```diff
@@ -17,7 +17,6 @@ all: lint install
 install: go.sum
 	go install -mod=readonly $(BUILD_FLAGS) ./cmd/bandd
 	go install -mod=readonly $(BUILD_FLAGS) ./cmd/bandcli
-	go install -mod=readonly $(BUILD_FLAGS) ./cmd/bandsv
```
smiled0g, line 20: this is no longer used?

chain/app/app.go
```diff
@@ -5,111 +5,103 @@ import (
...
+	"github.com/cosmos/cosmos-sdk/x/upgrade"
+	upgradeclient "github.com/cosmos/cosmos-sdk/x/upgrade/client"
 
 	bandsupply "github.com/bandprotocol/bandchain/chain/x/supply"
 	"github.com/bandprotocol/bandchain/chain/x/zoracle"
 )
 
 const (
-	appName          = "BandApp"
+	AppName          = "BandApp"
```
smiled0g, line 44: good change 👍

chain/app/app.go
```diff
@@ -120,246 +112,233 @@ type bandApp struct {
...
 func NewBandApp(
 	logger log.Logger, db dbm.DB, traceStore io.Writer, loadLatest bool,
-	invCheckPeriod uint, baseAppOptions ...func(*bam.BaseApp),
+	invCheckPeriod uint, skipUpgradeHeights map[int64]bool, home string,
+	baseAppOptions ...func(*bam.BaseApp),
 ) *bandApp {
 
-	// First define the top level codec that will be shared by the different modules
-	cdc := MakeCodec()
+	// TODO: Remove cdc in favor of appCodec once all modules are migrated.
```
smiled0g, line 153: so the `appCodec` is the new one that uses `Proto` instead of `amino`?

chain/app/db_app.go
```diff
@@ -52,15 +54,16 @@ func (app *dbBandApp) InitChain(req abci.RequestInitChain) abci.ResponseInitChai
 	app.cdc.MustUnmarshalJSON(req.AppStateBytes, &genesisState)
 
 	// Genaccount genesis
-	var genaccountsState genaccounts.GenesisState
-	genaccounts.ModuleCdc.MustUnmarshalJSON(genesisState[genaccounts.ModuleName], &genaccountsState)
+	// TODO: Figure out how the genesis account is populated
```
smiled0g: What does this mean? Would there be any difference in the genesis creation process and node mining #1 block?

chain/bandlib/client.go
```diff
@@ -107,7 +107,7 @@ func (client *BandStatefulClient) loop() {
 					{
 						client.msgs = append(client.msgs, msg.msg)
 						client.totalGas += msg.gas
-						client.totalFees = client.totalFees.Add(msg.fees)
+						client.totalFees = client.totalFees.Add(msg.fees...)
```
smiled0g, line 110: Why does this need to be changed?

chain/client/client.go
```diff
@@ -1,244 +1,244 @@
 package rpc
```
smiled0g, line 1: Will this file be revisited & ported later? I'd be nice if you've commented on why the entire file is commented and what's the TODO for this.

chain/client/proof.go
```diff
@@ -1,536 +1,536 @@
 package rpc
```
smiled0g, line 1: Same as above -- would be nice to have TODO here.

chain/client/proof_test.go
```diff
@@ -1,88 +1,88 @@
 package rpc
```
smiled0g, line 1: TODO needed.

chain/client/validators.go
```diff
@@ -1,64 +1,64 @@
 package rpc
```
smiled0g, line 1: TODO needed.

chain/cmd/bandcli/main.go
```diff
@@ -105,23 +116,46 @@ func txCmd(cdc *amino.Codec) *cobra.Command {
...
+		flags.LineBreak,
 		multiDelegateCommand(cdc),
-		client.LineBreak,
+		flags.LineBreak,
 	)
 
 	// add modules' tx commands
 	app.ModuleBasics.AddTxCommands(txCmd, cdc)
 
+	// remove auth and bank commands as they're mounted under the root tx command
```
smiled0g, line 134: Hmm .. why add `auth` and `bank` commands in the first place if we're going to remove it here? Doesn't make sense and looks like a hack to me.

chain/docker-config/add-account.exp
```diff
@@ -2,11 +2,7 @@ set user [lindex $argv 0]
 set pw [lindex $argv 1]
 set mnemonic [lindex $argv 2]
 
-spawn bandcli keys add $user --recover
-expect "Enter a passphrase to encrypt your key to disk:"
-send "$pw\r"
-expect "Repeat the passphrase:"
-send "$pw\r"
+spawn bandcli keys add $user --recover --keyring-backend test
```
smiled0g, line 5: Does this mean the passphrase is no longer needed? If so, then what if the user used a keystone from v0.37, would that be compatible with v0.38 here?

chain/docker-config/generate-genesis.sh
```diff
@@ -35,12 +35,11 @@ expect $DIR/add-account.exp \
     "smile stem oven genius cave resource better lunar nasty moon company ridge brass rather supply used horn three panic put venue analyst leader comic" \
 
 # add accounts to genesis
-bandd add-genesis-account $(bandcli keys show validator1 -a) 10000000000000uband
-bandd add-genesis-account $(bandcli keys show validator2 -a) 10000000000000uband
-bandd add-genesis-account $(bandcli keys show validator3 -a) 10000000000000uband
-bandd add-genesis-account $(bandcli keys show validator4 -a) 10000000000000uband
-
-bandd add-genesis-account $(bandcli keys show requester -a) 10000000000000uband
+bandd add-genesis-account validator1 10000000000000uband --keyring-backend test
```
smiled0g, line 38: what is the `test` here -- is it a flag/option or user-defined string that we hard-coded here?

chain/docker-config/single-validator/run.sh
```diff
@@ -13,4 +13,4 @@ mv config_tmp.toml config.toml
 cd /zoracle/
 
 mkdir ~/.banddb
-bandd start --rpc.laddr tcp://0.0.0.0:26657 --with-db sqlite3:/root/.banddb/main.db
+bandd start --rpc.laddr tcp://0.0.0.0:26657 --with-db "postgres: host=172.18.0.88 port=5432 user=postgres dbname=postgres password=postgrespassword sslmode=disable"
```
smiled0g, line 16: Let's add `# TODO` to remind us to remove the user & password. Not sure if it's configured to be accessible via localhost only, but it doesn't look secure here.

chain/go.mod
```diff
@@ -3,21 +3,30 @@ module github.com/bandprotocol/bandchain/chain
...
+	github.com/snikch/goodman v0.0.0-20171125024755-10e37e294daa
+	github.com/spf13/afero v1.2.2 // indirect
+	github.com/spf13/cobra v0.0.6
 	github.com/spf13/viper v1.6.2
 	github.com/stretchr/testify v1.5.1
 	github.com/tendermint/go-amino v0.15.1
-	github.com/tendermint/iavl v0.12.4
-	github.com/tendermint/tendermint v0.32.9
-	github.com/tendermint/tm-db v0.2.0
+	github.com/tendermint/tendermint v0.33.1-dev3
```
smiled0g: God bless `tendermint v0.33.1-dev3` 🙏👼

chain/x/zoracle/endblock.go
```diff
@@ -84,7 +84,7 @@ func handleEndBlock(ctx sdk.Context, keeper Keeper) sdk.Result {
 		// Must never overflow because we already checked for overflow above with
 		// gasConsumed + request.ExecuteGas (which is >= gasUsed).
 		if overflow {
-			panic(sdk.ErrorGasOverflow{Descriptor: "ExecuteRequest"})
+			panic("zoracle::handleEndBlock: Gas overflow")
```
smiled0g: Is `sdk.ErrorGasOverflow` deprecated? What's the reason for this change?

docker-compose.yaml
```diff
@@ -160,25 +158,6 @@ services:
     ports:
       - 80:80
 
-  code-server:
```
smiled0g, line 163: We no longer need this?

smiled0g (commented):
This is without the IBC right? I don't see new messages or changes under report/request handler.

EDIT: Nvm, just saw the change log.

chain/Makefile
sorawit: Yeah bandsv is not used now that the frontend can send transaction directly and there's a separate faucet process.

chain/app/db_app.go
sorawit: Yes. there are some changes. `genaccounts` module is deprecated and genesis accounts are written on `bank`, we will need to patch this code to work with that, but it's not critical

chain/bandlib/client.go
sorawit: They change the signature. The function now accepts variadic arguments (arbitrary number of Coin arguments) vs the old one that accepts one slice (one array of Coins).

chain/client/proof_test.go
sorawit: Sure.

chain/cmd/bandcli/main.go
sorawit: Not my code. It's from them! https://github.com/cosmos/gaia/blob/ibc-alpha/cmd/gaiacli/main.go#L134 

Didn't dig more, but I agree it looks hacky, but it must be here for some reasonss.

chain/docker-config/add-account.exp
sorawit: They change it so that there are multiple ways you can store your key. If user wants to use the same one as 0.37, they pass `--keyring-backend file`. The default option now uses the machine's keychain store.

chain/docker-config/generate-genesis.sh
sorawit: It's the configuration of where to store & encrypted key (mnemonic).

chain/docker-config/single-validator/run.sh
sorawit: This docker thing is primarily for testing.

chain/go.mod
sorawit: lol

chain/x/zoracle/endblock.go
sorawit: It's not. I did the change because I ran into some errors. Not sure why. But I change it back now and it works, so good catch!

chain/app/app.go
sorawit: It's a mess but yes. `appCodec` is their new codec effort.

docker-compose.yaml
sorawit: It's the `bandsv` thing. No longer needed.

chain/app/db_app.go
sorawit: From Cosmos change log:
Screen Shot 2020-04-05 at 00 50 34

chain/bandlib/client.go
sorawit: Screen Shot 2020-04-05 at 00 59 56

chain/app/db_app.go
sorawit: Nevermind. I patched it now. Tested locally and seemed to be working.

chain/app/genesis.go
```diff
@@ -5,119 +5,67 @@ import (
...
-// GenesisState defines a type alias for the Gaia genesis application state.
+// GenesisState defines a type alias for the Band genesis application state.
 type GenesisState map[string]json.RawMessage
 
 // NewDefaultGenesisState generates the default state for the application.
 func NewDefaultGenesisState() GenesisState {
+	cdc := codecstd.MakeCodec(ModuleBasics)
+	denom := "uband"
+
+	stakingGenesis := staking.DefaultGenesisState()
```
taobun, line 36: Prefer default instead of custom all fileds?

chain/db/validator.go
```diff
@@ -116,9 +116,14 @@ func (b *BandDB) handleMsgEditValidator(msg staking.MsgEditValidator) error {
 }
 
 func (b *BandDB) handleMsgCreateValidator(msg staking.MsgCreateValidator) error {
+	pubKey, err := sdk.GetPubKeyFromBech32(sdk.Bech32PubKeyTypeConsPub, msg.Pubkey)
```
taobun, line 119: Wow pubkey change to string ...

chain/x/supply/keeper.go
```diff
@@ -47,19 +49,24 @@ func (k WrappedSupplyKeeper) BurnCoins(
 	// Create the account if it doesn't yet exist.
 	acc := k.GetModuleAccount(ctx, moduleName)
 	if acc == nil {
-		return sdk.ErrUnknownAddress(fmt.Sprintf("module account %s does not exist", moduleName))
+		panic(sdkerrors.Wrapf(
```
taobun, line 52: Panic?

chain/x/zoracle/handler.go
```diff
@@ -6,11 +6,12 @@ import (
 	"github.com/bandprotocol/bandchain/chain/owasm"
 	"github.com/bandprotocol/bandchain/chain/x/zoracle/internal/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
 // NewHandler creates the msg handler of this module, as required by Cosmos-SDK standard.
 func NewHandler(keeper Keeper) sdk.Handler {
-	return func(ctx sdk.Context, msg sdk.Msg) sdk.Result {
+	return func(ctx sdk.Context, msg sdk.Msg) (*sdk.Result, error) {
```
taobun, line 14: Return pointer???

chain/x/zoracle/internal/keeper/querier_test.go
```diff
@@ -176,7 +176,7 @@ func TestQueryRequestById(t *testing.T) {
 		abci.RequestQuery{},
 	)
 	// It must return error request not found
-	require.Equal(t, types.CodeItemNotFound, err.Code())
+	require.NotNil(t, err)
```
taobun, line 179: Cannot test error type?

chain/app/app.go
```diff
@@ -395,6 +374,32 @@ func (app *bandApp) DeliverTx(req abci.RequestDeliverTx) (res abci.ResponseDeliv
...
+// InitChainer application update at chain initialization
+func (app *bandApp) InitChainer(ctx sdk.Context, req abci.RequestInitChain) abci.ResponseInitChain {
+	var genesisState simapp.GenesisState
+	app.cdc.MustUnmarshalJSON(req.AppStateBytes, &genesisState)
+
+	res := app.mm.InitGenesis(ctx, app.cdc, genesisState)
+
+	// Set Historical infos in InitChain to ignore genesis params
+	stakingParams := staking.DefaultParams()
+	stakingParams.HistoricalEntries = 1000
```
taobun, line 396: db app things? Have related logic in our chain?

chain/app/genesis.go
sorawit: Yes. I think this is safer. We overwrite exactly what we need and we leave other things with default values.

chain/db/validator.go
sorawit: Yes

chain/x/supply/keeper.go
sorawit: Yes. I'm making the behavior as similar to the original `BurnCoins` as possible.
https://github.com/cosmos/cosmos-sdk/blob/ibc-alpha/x/supply/keeper/bank.go#L130

chain/x/zoracle/handler.go
sorawit: Yes. The new spec of handler returns `(*sdk.Result, error)`

chain/x/zoracle/internal/keeper/querier_test.go
sorawit: Yes. I can confirm that ALL tests in Cosmos-SDK use this approach. No tests check for error code.

chain/app/app.go
sorawit: This is not db app things. It's from here https://github.com/cosmos/gaia/blob/ibc-alpha/app/app.go#L328. 

Now that I look at it again. It's `gaia`'s IBC hack to store old validator set. I think we should leave it here for now. We can move it to default genesis param later. We can talk more if you are still not convinced.

taobun (approved):

chain/cmd/bandd/init.go
```diff
@@ -162,17 +163,12 @@ func InitCmd(
...
-				},
-				Evidence: types.EvidenceParams{
-					MaxAge: 100000,
-				},
-				Validator: types.ValidatorParams{PubKeyTypes: []string{types.ABCIPubKeyTypeSecp256k1}},
-			}
+			genDoc.ConsensusParams = types.DefaultConsensusParams()
+			genDoc.ConsensusParams.Block.MaxBytes = 200000 // 0.2MB
+			genDoc.ConsensusParams.Block.MaxGas = 20000000 // 20M gas (Maximum oracle script size uses 15M)
+			genDoc.ConsensusParams.Block.TimeIotaMs = 1000 // 0.2MB
```
Benzbeeb, line 169: I think it should be 1s

Benzbeeb (changes requested):

sorawit:
It's getting real!

PR #1131, opened 2020-04-04
Implement loading animation for all cards

### LINKED ISSUE #1025: Implement loading animation for all cards
Should be fun!

prin-r:
Fix #1025


tansawit (dismissed):

scan/src/components/ChainInfoHighlights.re
```diff
@@ -50,36 +50,53 @@ module Styles = {
...
+      | _ => (ID.Block.ID(0), "")
+      };
+    let validators =
+      switch (validators_) {
+      | ApolloHooks.Subscription.Data(vs) => vs
+      | _ => [||]
+      };
 
     // TODO replace this Mock finance.
     let mockFinance: PriceHook.Price.t = {
```
sorawit: I don't think this is needed anymore? Let's have loading state if price is still not loaded.

scan/src/components/ChainInfoHighlights.re
```diff
@@ -50,36 +50,53 @@ module Styles = {
...
+    let loading =
+      switch (latestBlock_, validators_) {
+      | (ApolloHooks.Subscription.Data(_), ApolloHooks.Subscription.Data(_)) => false
+      | _ => true
+      };
 
-    let lastProcessedHeight = latestBlock.height;
-    let moniker = latestBlock.validator.moniker;
+    let (lastProcessedHeight, moniker) =
+      switch (latestBlock_) {
```
sorawit: This is too complicated and error prone. If we write it correct, we should never really need default dummy values.

sorawit (dismissed):

scan/src/components/ChainInfoHighlights.re
```diff
@@ -50,146 +50,142 @@ module Styles = {
...
+          {switch (valueAndExtraComponentSub) {
+           | Data(_) => <Text value=label color=Colors.bandBlue spacing={Text.Em(0.05)} />
+           | _ => <LoadingCensorBar width=90 height=18 />
+           }}
+        </div>
+        {switch (valueAndExtraComponentSub) {
+         | Data((valueComponent, _)) => valueComponent
+         | _ => <LoadingCensorBar width=120 height=20 />
+         }}
+        <div className={Styles.withWidth(170)}>
```
evilpeach: The div with style should send along with `extraComponent`.  No need to set it here.

evilpeach (changes requested):

scan/src/components/ChainInfoHighlights.re
```diff
@@ -50,146 +50,147 @@ module Styles = {
...
+           | _ => <LoadingCensorBar width=90 height=18 />
+           }}
         </div>
-        valueComponent
-        <div className={Styles.withWidth(170)}> extraComponent </div>
+        {switch (valueAndExtraComponentSub) {
+         | Data((valueComponent, extraComponent)) => <> valueComponent extraComponent </>
+         | _ =>
+           <> <LoadingCensorBar width=120 height=20 /> <LoadingCensorBar width=75 height=15 /> </>
+         }}
```
evilpeach, line 70: IMO, we can put the switch `valueAndExtraComponentSub ` together and then switch on `bgUrl` inside.

evilpeach (changes requested):

evilpeach (approved):

scan/src/components/ChainInfoHighlights.re
evilpeach: However, I think it's ok for now.

tansawit (approved):

PR #1154, opened 2020-04-07
Implement POC for IBC oracle request and response packet

sorawit:


chain/x/zoracle/handler.go
```diff
@@ -23,13 +25,60 @@ func NewHandler(keeper Keeper) sdk.Handler {
...
+			// return &sdk.Result{Events: ctx.EventManager().Events().ToABCIEvents()}, nil
 			return handleMsgRequestData(ctx, keeper, msg)
 		case MsgReportData:
 			return handleMsgReportData(ctx, keeper, msg)
 		case MsgAddOracleAddress:
 			return handleMsgAddOracleAddress(ctx, keeper, msg)
 		case MsgRemoveOracleAddress:
 			return handleMsgRemoveOracleAddress(ctx, keeper, msg)
+		case channeltypes.MsgPacket:
+			var requestData OracleRequestPacketData
```
taobun, line 69: Why not use the same pattern
```
err := ...
if err != nil {
    return err...
}
```

taobun (changes requested):

chain/x/zoracle/handler.go
sorawit: Because it's possible to try multiple types of packets?

taobun (approved):

PR #1176, opened 2020-04-09
Make expiration time a module parameter

### LINKED ISSUE #1081: Make expiration time a module parameter
Now that we have timeout for lambda execution, I think it makes sense to force everything have the same exp. This also allows us to figure out which reporter fails to report data in the future. Let's have 20 blocks as the default value.

Update everything that includes expiration time to not have it.

Benzbeeb:
Fixed #1081


chain/db/request.go
```diff
@@ -4,6 +4,7 @@ import (
 	"strconv"
 
 	"github.com/bandprotocol/bandchain/chain/x/oracle"
+	"github.com/bandprotocol/bandchain/chain/x/oracle/types"
```
taobun: Remove this line

chain/db/request.go
```diff
@@ -169,12 +170,16 @@ func (b *BandDB) handleMsgRequestData(
 	if err != nil {
 		return err
 	}
+	request, err := b.OracleKeeper.GetRequest(b.ctx, types.RequestID(id))
```
taobun: Use `oracle.RequestID`

taobun: but we have 2 place to add new request (ibc.go), so in my opinion change this on `AddNewRequest` function instead

taobun (changes requested):

sorawit (dismissed):
Thank you!

sorawit (approved):

PR #1181, opened 2020-04-09
scan: Make LatestBlocks and ChainInfoHighlights share same block info

sorawit:
This to make sure last block info is always updated together in both components.


sorawit:
@tansawit @evilpeach @prin-r An example of how you can pass `Sub` around to child components to make sure each component can have separate loading states while sharing the same data source.

evilpeach (approved):
Thanks!

prin-r (approved):

PR #1189, opened 2020-04-10
Add median and test with average

### LINKED ISSUE #1114: Implement owasm load_median function
Should work for any type `T` that has `FromStr` and `PartialOrd`. Also write tests!

taobun:
fixed #1114


owasm/src/ext/mod.rs
```diff
@@ -29,3 +40,80 @@ where
     // TODO
     oei::get_external_data(external_id, 0).unwrap().parse::<T>().ok().unwrap()
 }
+
+pub fn load_median<T>(external_id: i64) -> Option<T>
+where
+    T: std::str::FromStr + std::cmp::PartialOrd + std::marker::Copy + num::Num + num::NumCast,
+{
+    let mut vals: Vec<T> = Vec::new();
```
sorawit: Let's have a separate function to do this loading for both `load_average` and `load_median`

sorawit (changes requested):

sorawit (approved):
OK for now. There must be a way to do it without requiring this much traits, but let's save that for another day.

Benzbeeb (approved):

PR #1200, opened 2020-04-11
owasm: Add Pythonic CoinGecko scripts that take symbol as argument

sorawit:


chain/datasources/coingecko_price.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
```
sorawit, line 1: This has been tested to work with the current lambda release.

chain/datasources/coingecko_price.py
```diff
@@ -0,0 +1,30 @@
...
+import json
+import urllib.request
+import sys
+
+COINS_URL = "https://api.coingecko.com/api/v3/coins/list"
+PRICE_URL = "https://api.coingecko.com/api/v3/simple/price?ids={}&vs_currencies=usd"
+
+
+def make_json_request(url):
+    return json.loads(urllib.request.urlopen(url).read())
```
sorawit, line 12: We don't use `requests` here since we don't install that on lambda package yet.

taobun (approved):

sorawit:
@evilpeach @taobun  Please note that frontend will need to be able to parse Python script in execute dryrun.

Benzbeeb (approved):

PR #1225, opened 2020-04-11
chain/scan: Remove prepare/execute gas from request parameter

sorawit:
In the effort to simplify user experience, we plan to remove the need for users to specify `prepareGas` and `executeGas`. This PR removes prepare gas from all places and simply hard-code 100k gas for now. Note that this PR also updates scan code.


chain/x/oracle/handler_test.go
```diff
@@ -508,295 +508,7 @@ func TestEndBlock(t *testing.T) {
 }
 
 func TestEndBlockExecuteFailedIfExecuteGasLessThanGasUsed(t *testing.T) {
```
sorawit, line 510: These tests are removed as we are moving toward simpler EndBlock logic. I will make sure we have good tests later!

chain/x/oracle/keeper/request.go
```diff
@@ -54,7 +54,8 @@ func (k Keeper) AddRequest(
 		validators[i] = validatorsByPower[i].GetOperator()
 	}
 
-	ctx.GasMeter().ConsumeGas(executeGas, "ExecuteGas")
+	// TODO: Remove this hardcode (and KeyEndBlockExecuteGasLimit param)!
+	executeGas := uint64(100000)
```
taobun, line 58: User doesn't need to pay execute gas. WOW

sorawit: Yep. It's all part of the package!

taobun (approved):

PR #1230, opened 2020-04-12
scan: Make subscription query more concise and conventional

sorawit:


sorawit:
@prin-r Please note this. It's all your code.

scan/src/subscriptions/RequestSub.re
```diff
@@ -74,6 +53,31 @@ module Mini = {
...
+      subscription RequestsMiniByOracleScript($id: bigint!, $limit: Int!, $offset: Int!) {
+        requests(
+          where: {oracle_script_id: $id}
+          limit: $limit
+          offset: $offset
+          order_by: {id: desc}
+        ) {
+          id @bsDecoder(fn: "ID.Request.fromJson")
+          requester @bsDecoder(fn: "Address.fromBech32")
+          oracle_script @bsRecord {
```
evilpeach, line 67: Should it be Camel case?

evilpeach (changes requested):

scan/src/subscriptions/RequestSub.re
sorawit: This is old code. Otherwise I will need to change the record definition too. If it needs to change, then it should go to a separate PR.

prin-r (approved):

evilpeach (approved):

PR #1254, opened 2020-04-13
Add gold price data source

Benzbeeb:


sorawit (dismissed):
Looks great

chain/datasources/binance_atom_price.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
```
taobun: Why we need this script? Using old binance script?

owasm/chaintests/gold_price/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+struct Output {
+    px: u64,
+}
+
+#[no_mangle]
+fn prepare_impl(input: Input) {
+    // Gold price data source
+    oei::request_external_data(5, 1, input.symbol.as_bytes());
+    // Binance data source
+    oei::request_external_data(6, 2, input.symbol.as_bytes());
```
taobun: I think we can use existing Binance script and send "ATOM".as_bytes() here instead

owasm/chaintests/gold_price/src/lib.rs
```diff
@@ -0,0 +1,31 @@
+use borsh::{BorshDeserialize, BorshSerialize};
+use owasm::{execute_entry_point, ext, oei, prepare_entry_point};
+
+#[derive(BorshDeserialize)]
+struct Input {
+    symbol: String,
```
taobun: Don't need it in this case

taobun (changes requested):

owasm/chaintests/gold_price/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+
+#[derive(BorshSerialize)]
+struct Output {
+    px: u64,
+}
+
+#[no_mangle]
+fn prepare_impl(input: Input) {
+    // Gold price data source
+    oei::request_external_data(5, 1, input.symbol.as_bytes());
```
taobun: Help me test if we send empty bytes, it will work correctly

taobun (changes requested):

chain/app/genesis.go
```diff
@@ -131,6 +136,11 @@ func GetDefaultDataSourcesAndOracleScripts(owner sdk.AccAddress) json.RawMessage
 			"Oracle script for getting an average crypto price from many sources encoding parameter by borsh.",
 			"./owasm/res/crypto_price_borsh.wasm",
 		},
+		{
+			"Gold price script",
+			"Oracle script for getting an average gold price from many sources.",
```
taobun: Many sources LUL

taobun: Oracle script for getting an average gold price in ATOM

taobun (changes requested):

taobun (approved):

PR #1257, opened 2020-04-13
Bring extra fields in tx to camelCase

taobun:
We will patch these field to snake case after patch all msgs to protobuf


taobun:
Fixed by front end already

PR #1273, opened 2020-04-15
Scan: Binding ledger-cosmos-js

### LINKED ISSUE #1203: Implement binding for ledger-cosmos-js in Scan
Binding [Ledger cosmos js](https://github.com/cosmos/ledger-cosmos-js) on Scan with example code

evilpeach:
Fixes #1203 

- [x] Binding `ledger-cosmos-js`
- [x] Send Request tx via ledger
- [x] Implement good abstraction


scan/src/bindings/Ledger.re
```diff
@@ -0,0 +1,70 @@
...
+
+let getAddress = () => {
+  getApp()
+  |> Js.Promise.then_(app => {
+       //  TODO: remove hard-coded path later
+       let path = [|44, 118, 4, 0, 2|];
+       let prefix = "cosmos";
+       let responsePromise = getAddressAndPubKey(app, path, prefix);
+       let%Promise response = responsePromise;
+       Js.Console.log(response.bech32_address);
```
sorawit: Why do we log?

sorawit (changes requested):

scan/src/wallet/Wallet.re
```diff
@@ -0,0 +1,24 @@
+type t =
+  | Mnemonic(Mnemonic.t)
+  | Ledger(Ledger.t);
+
+let createFromMnemonic = (rpcUrl, mnemonic) => {
+  let bandChain = BandWeb3.network(rpcUrl, "bandchain");
```
sorawit: I think we can put arbitrary url and chain id here. Reading https://github.com/cosmostation/cosmosjs/blob/master/src/index.js, I don't think it requires url or chain id for getting address or signing.

scan/src/wallet/Wallet.re
```diff
@@ -0,0 +1,24 @@
...
+  bandChain->BandWeb3.setPath("m/44'/494'/0'/0/0");
+  bandChain->BandWeb3.setBech32MainPrefix("band");
+  Mnemonic({bandChain, mnemonic});
+};
+
+let createFromLedger = path => {
```
sorawit: I think this should be a promise that errors if a ledger is not connected?

sorawit (changes requested):

evilpeach:
@sorawit Currently, there are snake case variables on the binding. Is there a way to handle this properly instead of creating `raw_response_t`(contains snake case) and `response_t` (camel case).

sorawit (approved):

tansawit (approved):

PR #1390, opened 2020-04-22
chain: Revive get proof endpoint

taobun:
- Refactor proof struct to include request and response packet
- Add IAVL + packets to proof
- Refactor TMSignature code
- Bring `/bandchain/proof/{requestID} to work again
- Comment EVM proof part will be implemented after getting ABI of a new contract.
- Can combine both packets to result_hash but in solidity need to research how to encode protobuf type to bytes in solidity (I already researched about it need to make it completed).
- Signatures haven't verified. It will have pr to handle it.
- Revisit GetMultiStore struct because I miss capability module.


Benzbeeb (approved):

sorawit (approved):
This is good.

PR #1475, opened 2020-05-01
chain: Recover panic from VM.Run 

### LINKED ISSUE #1377: owasm::Execute should return error instead of panic if vm.Run panics
Apparently vm.Run can panic in many cases, including when param count mismatch. We don't want to crash the chain during EndBlock when that happens.

taobun:
fixed #1377 

This is a result when request on bad oracle script
```
{
  "jsonrpc": "2.0",
  "id": -1,
  "result": {
    "hash": "394D9DD256F17D6C4234F4B118AC626C78624A0BE84B794AA5D27D9C88B9D6F5",
    "height": "39",
    "index": 0,
    "tx_result": {
      "code": 7,
      "data": null,
      "log": "failed to execute message; message index: 0: param count mismatch: bad wasm execution",
      "info": "",
      "gasWanted": "1000000",
      "gasUsed": "419897",
      "events": [],
      "codespace": "oracle"
    },
    "tx": "KCgWqQoqQnCONAgDEgEDGAEgASoFdGVzdDIyFN0+AvKRiDpO1BGaSf1z4gWDa6MlEgQQwIQ9GmoKJuta6YchA6rU4D6xY4ibvdSDdg8iLDSVF73O6ZB/vUH6iJsNRNqYEkDXFTv9NkZvIzK41s8jE5llt5JFf43PPwVTSWHBp8R0+kFcTy4UPnr/wb8C3l6oQGCoJwVSYjubWccOUnxcivGj"
  }
}
```

### DIFF
--- a/chain/owasm/execute.go
+++ b/chain/owasm/execute.go
@@ -1,6 +1,7 @@
 package owasm
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/perlin-network/life/exec"
@@ -36,6 +37,20 @@ func Execute(
 	if !ok {
 		return nil, 0, fmt.Errorf("Execute: invalid owasm entry: %s", entry)
 	}
+	defer func() {
+		if r := recover(); r != nil {
+			switch x := r.(type) {
+			case string:
+				err = errors.New(x)
+			case error:
+				err = x
+			default:
+				err = errors.New("Unknown panic")
+			}
+			result = nil
+			gasUsed = 0
+		}
+	}()
 	_, err = vm.Run(int(entryID))
 	return resolver.result, vm.Gas, err
 }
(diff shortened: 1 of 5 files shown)


chain/owasm/execute.go
```diff
@@ -36,6 +37,20 @@ func Execute(
...
 	}
+	defer func() {
+		if r := recover(); r != nil {
+			switch x := r.(type) {
+			case string:
+				err = errors.New(x)
+			case error:
+				err = x
+			default:
+				err = errors.New("Unknown panic")
```
sorawit, line 48: Let's print the 'r' out. so if this happens we know what's going on.

sorawit (approved):
Looks good

Benzbeeb (approved):

Issue #1239, opened 2020-04-12
Find a Pseudo-randomly way to choose validators to perform oracle tasks


sorawit:
Instead of relying on top validators, the approach the makes more sense would be to randomly choose validators (chances based on staking power) to perform tasks. We want to research what method used to select validator is the most make sense. Experiment and report to @sorawit

PR #1476, opened 2020-05-01
Pseudo-randomly choose validators to perform oracle tasks

Benzbeeb:
#1239


taobun:
This pr doesn't aim to random validator in our chain code?

PR #1472, opened 2020-05-01
scan: fix transaction signature and send message amount type

tansawit:

PR #1477, opened 2020-05-01
Scan: Remove chainID hardcoded

evilpeach:
Review this PR after #1472 has merged.

- Remove hardcoded in `createRawTx`

### DIFF
--- a/scan/src/components/UserAccount.re
+++ b/scan/src/components/UserAccount.re
@@ -179,30 +179,35 @@ module Balance = {
 
 [@react.component]
 let make = () => {
+  let metadataSub = MetadataSub.use();
   let (accountOpt, dispatchAccount) = React.useContext(AccountContext.context);
   let (_, dispatchModal) = React.useContext(ModalContext.context);
 
-  let connect = () => dispatchModal(OpenModal(Connect));
+  let connect = chainID => dispatchModal(OpenModal(Connect(chainID)));
   let disconnect = () => dispatchAccount(Disconnect);
   let submitTx = () => dispatchModal(OpenModal(SubmitTx));
 
-  switch (accountOpt) {
-  | Some({address}) =>
-    <>
-      <Row justify=Row.Right>
-        <Col> <AddressRender address position=AddressRender.Nav /> </Col>
-        <Col> <HSpacing size={`px(27)} /> </Col>
-        <Col> <DisconnectBtn disconnect /> </Col>
-      </Row>
-      <VSpacing size=Spacing.md />
-      <Row justify=Row.Right>
-        <Col> <Balance address /> </Col>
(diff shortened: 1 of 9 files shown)


tansawit (changes requested):
Right now there is a loading state every time the green connect button is pressed (because the chainID is being retrieved every time). Ideally the chainID should only be fetched once during the first connect button press.

sorawit (changes requested):
Agree with @tansawit . Let's make the green connect passes around `ChainID` and not load it in the modal components.

scan/src/components/UserAccount.re
```diff
@@ -179,30 +179,35 @@ module Balance = {
...
-        <Col> <Balance address /> </Col>
-        <Col> <HSpacing size={`px(5)} /> </Col>
-        <Col> <FaucetBtn address={address->Address.toBech32} /> </Col>
-      </Row>
-      <VSpacing size=Spacing.md />
-      <Row justify=Row.Right> <Col> <SubmitTxBtn submitTx /> </Col> </Row>
-    </>
-  | None => <Col> <ConnectBtn connect /> </Col>
+  switch (metadataSub) {
+  | Data({chainID}) =>
```
sorawit: Why don't we switch on `accountOpt` first? So we always show disconnect button even when we don't have chainID?

scan/src/components/UserAccount.re
```diff
@@ -179,30 +179,35 @@ module Balance = {
...
+          <Col> <Balance address /> </Col>
+          <Col> <HSpacing size={`px(5)} /> </Col>
+          <Col> <FaucetBtn address={address->Address.toBech32} /> </Col>
+        </Row>
+        <VSpacing size=Spacing.md />
+        <Row justify=Row.Right> <Col> <SubmitTxBtn submitTx /> </Col> </Row>
+      </>
+    | None => <Col> <ConnectBtn connect={_ => connect(chainID)} /> </Col>
+    }
+  | _ => React.null
```
sorawit: Can we do the loading animation in this PR? How hard is it?

sorawit (changes requested):

scan/src/components/UserAccount.re
evilpeach: If we don't have chainID, we show nothing.

scan/src/components/UserAccount.re
evilpeach: No, I'll add it.

scan/src/components/UserAccount.re
evilpeach: Oh, my bad. just push edited version.

sorawit (approved):

tansawit (approved):

PR #1513, opened 2020-05-04
chain: clean up oracle module code

sorawit:
Also make sure that OnRecvPacket return errors on unknown packs.

### DIFF
--- a/chain/x/oracle/module.go
+++ b/chain/x/oracle/module.go
@@ -12,10 +12,12 @@ import (
 	"github.com/bandprotocol/bandchain/chain/x/oracle/types"
 	"github.com/cosmos/cosmos-sdk/codec"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/cosmos/cosmos-sdk/types/module"
 	"github.com/cosmos/cosmos-sdk/x/capability"
 	channel "github.com/cosmos/cosmos-sdk/x/ibc/04-channel"
 	channelexported "github.com/cosmos/cosmos-sdk/x/ibc/04-channel/exported"
 	channeltypes "github.com/cosmos/cosmos-sdk/x/ibc/04-channel/types"
+	port "github.com/cosmos/cosmos-sdk/x/ibc/05-port"
 	ibctypes "github.com/cosmos/cosmos-sdk/x/ibc/types"
 
 	"github.com/cosmos/cosmos-sdk/client/context"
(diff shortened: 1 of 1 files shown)


Benzbeeb (approved):

chain/x/oracle/module.go
```diff
@@ -49,199 +56,149 @@ func (AppModuleBasic) ValidateGenesis(cdc codec.JSONMarshaler, bz json.RawMessag
...
-			requestData.MinCount, requestData.ClientID,
-			sdk.AccAddress([]byte("Unknown")),
-		)
-		return handleMsgRequestDataIBC(
-			ctx, am.keeper, newMsg, packet.GetDestPort(), packet.GetDestChannel(),
-		)
+// OnRecvPacket implements ics-26 IBCModule interface.
+func (am AppModule) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet) (*sdk.Result, error) {
+	var req OracleRequestPacketData
+	if err := types.ModuleCdc.UnmarshalJSON(packet.GetData(), &req); err != nil {
```
sorawit, line 177: Note this is the only non-trivial change. I make it return error if message parsing fails.

chain/x/oracle/module.go
```diff
@@ -12,33 +12,40 @@ import (
...
 	ibctypes "github.com/cosmos/cosmos-sdk/x/ibc/types"
 
 	"github.com/cosmos/cosmos-sdk/client/context"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	abci "github.com/tendermint/tendermint/abci/types"
 )
 
-// AppModule Basics object
+var (
+	_ module.AppModule      = AppModule{}
```
taobun, line 29: Cosmos-sdk write these line. But why we need these?

chain/x/oracle/module.go
```diff
@@ -49,199 +56,149 @@ func (AppModuleBasic) ValidateGenesis(cdc codec.JSONMarshaler, bz json.RawMessag
...
+	counterparty channeltypes.Counterparty, version string,
 ) error {
-	// // TODO: Enforce ordering, currently relayers use ORDERED channels
-	// if version != types.Version {
-	// 	return sdkerrors.Wrapf(porttypes.ErrInvalidPort, "invalid version: %s, expected %s", version, "ics20-1")
-	// }
-
-	// Claim channel capability passed back by IBC module
-	if err := am.keeper.ScopedKeeper.ClaimCapability(ctx, chanCap, ibctypes.ChannelCapabilityPath(portID, channelID)); err != nil {
+	// Claim channel capability passed back by IBC module.
```
taobun, line 131: Remove version?

chain/x/oracle/module.go
```diff
@@ -49,199 +56,149 @@ func (AppModuleBasic) ValidateGenesis(cdc codec.JSONMarshaler, bz json.RawMessag
...
-		newMsg := NewMsgRequestData(
-			requestData.OracleScriptID, calldata, requestData.AskCount,
-			requestData.MinCount, requestData.ClientID,
-			sdk.AccAddress([]byte("Unknown")),
-		)
-		return handleMsgRequestDataIBC(
-			ctx, am.keeper, newMsg, packet.GetDestPort(), packet.GetDestChannel(),
-		)
+// OnRecvPacket implements ics-26 IBCModule interface.
+func (am AppModule) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet) (*sdk.Result, error) {
```
taobun, line 175: We need to cal packet executed?

taobun (changes requested):

chain/x/oracle/module.go
sorawit: It's not strictly necessary, but it makes compile error messages easier to follow.

chain/x/oracle/module.go
sorawit: Yeah I don't think it's needed for us. Let's just skip it.

chain/x/oracle/module.go
sorawit: Don't think so. That will require we also define ack packet. Let's just say for now we don't follow ICS ack standard.

taobun (approved):

PR #1515, opened 2020-05-05
chain: Refactor SendOracleResponse out and improve request ibc flow

sorawit:
- Make SendOracleResponse not depend on request keeper at all. It just sends packets.
- Make sure expired requests get sent back to requester.
- Use error as result for failed requests.
- Merge NewRequest function into one.

### DIFF
--- a/chain/x/oracle/endblock.go
+++ b/chain/x/oracle/endblock.go
@@ -1,6 +1,8 @@
 package oracle
 
 import (
+	"fmt"
+
 	"github.com/bandprotocol/bandchain/chain/x/oracle/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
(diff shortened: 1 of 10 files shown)


chain/x/oracle/keeper/request.go
```diff
@@ -78,110 +76,60 @@ func (k Keeper) AddRequest(ctx sdk.Context, req types.Request) (types.RequestID,
 	return id, nil
 }
 
-func (k Keeper) resolveRequest(
-	ctx sdk.Context, reqID types.RequestID, resolveStatus types.ResolveStatus, result []byte,
+func (k Keeper) ResolveRequest(
```
taobun, line 81: Can you add a comment to make this function more understandable?

chain/x/oracle/endblock.go
```diff
@@ -14,13 +14,16 @@ func resolveRequest(ctx sdk.Context, k Keeper, reqID types.RequestID) {
 	script := k.MustGetOracleScript(ctx, req.OracleScriptID)
 	executeGas := k.GetParam(ctx, KeyExecuteGas)
 	result, _, err := k.OwasmExecute(env, script.Code, "execute", req.Calldata, executeGas)
-	var resolveStatus types.ResolveStatus
+
+	var res types.OracleResponsePacketData
 	if err != nil {
-		resolveStatus = types.Failure
+		res = k.ResolveRequest(ctx, reqID, types.Failure, []byte(err.Error()))
```
taobun: Error has been used?

taobun (changes requested):

chain/x/oracle/endblock.go
sorawit: Actually error should not be part of the consensus. I do want to allow Owasm to emit errors somehow though, but will save that for other diffs.

chain/x/oracle/keeper/request.go
sorawit: Yes can do.

taobun (approved):

Benzbeeb (approved):

PR #1651, opened 2020-05-16
chain: Remove unnecessary ibc packet code in the current version

sorawit:


sorawit:
@Benzbeeb Please help verify if this does not break IBC

PR #1667, opened 2020-05-18
chain: Add request and report to genesis

perimeko:


sorawit:
We talked. Let's close this for now.

PR #1676, opened 2020-05-19
chain: Add result to genesis

perimeko:


chain/x/oracle/keeper/result_test.go
```diff
@@ -69,3 +78,60 @@ package keeper
 // 	err = keeper.AddResult(ctx, 1, 1, []byte("calldata"), []byte("result"))
 // 	require.NotNil(t, err)
 // }
+
+func TestGetAllResults(t *testing.T) {
+	_, ctx, k := createTestInput()
+
+	reqPacket1 := oracle.OracleRequestPacketData{
```
sorawit: Let's use New factory function (`NewOracleRequestPacketData`)

chain/x/oracle/keeper/result_test.go
```diff
@@ -69,3 +78,60 @@ package keeper
...
+		Result:        BasicCalldata,
+	}
+
+	_, err = k.AddResult(ctx, types.RequestID(4), reqPacket4, resPacket4)
+	require.NoError(t, err)
+
+	results := k.GetAllResults(ctx)
+
+	require.Equal(t, 4, len(results))
+	require.NotEmpty(t, results[0])
```
sorawit: Let's check for the exact value.

chain/x/oracle/keeper/result.go
```diff
@@ -50,3 +50,24 @@ func (k Keeper) AddResult(
 	store.Set(types.ResultStoreKey(id), resultHash)
 	return resultHash, nil
 }
+
+// GetAllResults return the list of all results in the store, or nil if there is none
```
sorawit: GetAllResults returns the list of all results in the store. Nil will be added for skipped results.

sorawit (changes requested):

chain/x/oracle/genesis.go
```diff
@@ -73,6 +77,10 @@ func InitGenesis(ctx sdk.Context, k Keeper, data GenesisState) []abci.ValidatorU
 		k.SetValidatorReportInfo(ctx, info.Validator, info)
 	}
 
+	for i, result := range data.Results {
+		k.SetResult(ctx, types.RequestID(i+1), result)
```
sorawit: Let's skip if result is nil

sorawit (changes requested):

chain/x/oracle/genesis.go
perimeko: fixed

chain/x/oracle/keeper/result_test.go
perimeko: fixed

chain/x/oracle/keeper/result.go
```diff
@@ -50,3 +50,30 @@ func (k Keeper) AddResult(
 	store.Set(types.ResultStoreKey(id), resultHash)
 	return resultHash, nil
 }
+
+// SetResult set result to the store.
+func (k Keeper) SetResult(ctx sdk.Context, regID types.RequestID, result []byte) {
```
sorawit: Move it to above `AddResult`. Also please name variable `reqID`

chain/x/oracle/keeper/result.go
```diff
@@ -50,3 +50,30 @@ func (k Keeper) AddResult(
 	store.Set(types.ResultStoreKey(id), resultHash)
```
sorawit: Now that we have `SetResult`. I think this should call into `SetResult`.

chain/x/oracle/keeper/result_test.go
```diff
@@ -69,3 +76,47 @@ package keeper
...
+	require.Equal(t, resultHashReqID4, results[3])
+}
+
+func TestSetResult(t *testing.T) {
+	_, ctx, k := createTestInput()
+
+	reqPacket := types.NewOracleRequestPacketData("alice", 1, BasicCalldata, 1, 1)
+	resPacket := types.NewOracleResponsePacketData("alice", 1, 1, 1589535020, 1589535022, 1, BasicCalldata)
+
+	resultHash, err := k.AddResult(ctx, types.RequestID(1), reqPacket, resPacket)
```
sorawit, line 113: Let's also test `GetResult` here

chain/x/oracle/keeper/result_test.go
```diff
@@ -69,3 +76,47 @@ package keeper
...
+
+	resultHashReqID4, err := k.AddResult(ctx, types.RequestID(4), reqPacket4, resPacket4)
+	require.NoError(t, err)
+
+	results := k.GetAllResults(ctx)
+
+	require.Equal(t, 4, len(results))
+	require.Equal(t, resultHashReqID1, results[0])
+
+	// result of reqID 2 and 3 should be empty byte array
```
sorawit: It should be nil, not empty byte array

chain/x/oracle/keeper/result.go
```diff
@@ -50,3 +50,30 @@ func (k Keeper) AddResult(
...
+
+// GetAllResults returns the list of all results in the store. Nil will be added for skipped results.
+func (k Keeper) GetAllResults(ctx sdk.Context) (results [][]byte) {
+	store := ctx.KVStore(k.storeKey)
+	iterator := sdk.KVStorePrefixIterator(store, types.ResultStoreKeyPrefix)
+	var previousReqID types.RequestID
+	defer iterator.Close()
+	for ; iterator.Valid(); iterator.Next() {
+		currentReqID := types.RequestID(sdk.BigEndianToUint64(iterator.Key()[1:]))
+		diffReqIDCount := currentReqID - previousReqID
```
sorawit: let's use `:= int(...)` here then we don't need to do casting in the for statement.

chain/x/oracle/keeper/result.go
```diff
@@ -50,3 +50,30 @@ func (k Keeper) AddResult(
...
+
+		// Insert nil for the request that result is empty
```
sorawit: Insert nil for each request without result.

sorawit (changes requested):

chain/x/oracle/keeper/result.go
```diff
@@ -50,3 +50,30 @@ func (k Keeper) AddResult(
 	store.Set(types.ResultStoreKey(id), resultHash)
 	return resultHash, nil
 }
+
+// SetResult set result to the store.
+func (k Keeper) SetResult(ctx sdk.Context, reqID types.RequestID, result []byte) {
```
sorawit: Let's move this function to above `GetResult` for consistency.

sorawit (changes requested):

chain/x/oracle/keeper/result.go
perimeko: fixed

chain/x/oracle/keeper/result.go
perimeko: fixed, used set result function

chain/x/oracle/keeper/result_test.go
perimeko: fixed

chain/x/oracle/keeper/result.go
perimeko: fixed

sorawit (approved):

chain/x/oracle/keeper/result.go
```diff
@@ -13,6 +13,12 @@ func (k Keeper) HasResult(ctx sdk.Context, id types.RequestID) bool {
 	return ctx.KVStore(k.storeKey).Has(types.ResultStoreKey(id))
 }
 
+// SetResult set result to the store.
```
sorawit: sets

sorawit (changes requested):

sorawit (approved):

taobun (approved):

PR #1677, opened 2020-05-19
Scan: Polish for wenchang

evilpeach:
- Edit typo
- Adjust digits
- Change tx mode from `block` to `sync` for fixing request timeout

### DIFF
--- a/scan/src/components/modal/submitTx/PreviewJsonStep.re
+++ b/scan/src/components/modal/submitTx/PreviewJsonStep.re
@@ -119,7 +119,8 @@ let make = (~rawTx, ~onBack, ~account: AccountContext.t) => {
                         ~signature=signature |> JsBuffer.toBase64,
                         ~pubKey=account.pubKey,
                         ~tx=rawTx,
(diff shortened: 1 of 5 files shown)


tansawit (dismissed):

scan/src/components/modal/submitTx/PreviewJsonStep.re
```diff
@@ -119,7 +119,8 @@ let make = (~rawTx, ~onBack, ~account: AccountContext.t) => {
                         ~signature=signature |> JsBuffer.toBase64,
                         ~pubKey=account.pubKey,
                         ~tx=rawTx,
-                        ~mode="block",
+                        // TODO: Use sync for now
```
prin-r: Should say what to do next

evilpeach: We use sync mode instead of block permanently.

prin-r (approved):

tansawit (approved):

PR #1682, opened 2020-05-20
chain: Make sure to close iterator

sorawit:


obi/example.long.obi
```diff
@@ -0,0 +1,6 @@
+{
```
taobun: What?

taobun (changes requested):

PR #1686, opened 2020-05-20
oracled2: Handle ibc packet

perimeko:


chain/x/oracle/types/events.go
```diff
@@ -12,6 +12,7 @@ const (
 	EventTypeAddReporter        = "add_reporter"
 	EventTypeRemoveReporter     = "remove_reporter"
 	EventTypeRequestExecute     = "request_execute"
+	EventTypeMessage            = "message"
```
sorawit: Let's not define new variables. Search for it in cosmos-sdk repo. For example:

```
➜  cosmos-sdk git:(master) ✗ git grep '= \"message\"' | cat
types/events.go:	EventTypeMessage = "message"
```

chain/cmd/bandoracled2/handler.go
```diff
@@ -26,16 +26,25 @@ func handleTransaction(c *Context, l *Logger, tx tmtypes.TxResult) {
...
 			continue
 		}
-		if messageType != "request" {
+
+		if messageType == otypes.MessageTypeRequest {
+			go handleRequestLog(c, l, log)
+		} else if messageType == otypes.MessageTypeOraclePacket {
+			// Try to get request id from packet. If not then return error.
+			_, err := GetEventValue(log, otypes.EventTypeRequest, otypes.AttributeKeyID)
+			if err != nil {
```
sorawit, line 40: If err not nil, that means it's not an oracle packet, which is totally ok. I think we should debug log that we skip non-request packet.

chain/cmd/bandoracled2/handler.go
```diff
@@ -26,16 +26,25 @@ func handleTransaction(c *Context, l *Logger, tx tmtypes.TxResult) {
...
+				l.Error(":cold_sweat: Failed to parse request id from oracle packet with error: %s", err.Error())
+				return
+			}
+			go handleRequestLog(c, l, log)
+		} else {
 			l.Debug(":ghost: Skipping non-request message type: %s", messageType)
```
sorawit: We should update this log to skipping non-{request/packet} message type

sorawit (changes requested):

chain/x/oracle/types/events.go
```diff
@@ -30,4 +31,8 @@ const (
 	AttributeKeyResolveStatus  = "resolve_status"
 	AttributeKeyResult         = "result"
 	AttributeKeyResultHash     = "result_hash"
+	AttributeKeyAction         = "action"
+
+	MessageTypeRequest      = "request"
+	MessageTypeOraclePacket = "ics04/opaque"
```
taobun: I think we don't need to declare these variable

taobun (changes requested):

chain/cmd/bandoracled2/handler.go
```diff
@@ -25,17 +26,25 @@ func handleTransaction(c *Context, l *Logger, tx tmtypes.TxResult) {
...
+		messageType, err := GetEventValue(log, sdk.EventTypeMessage, otypes.AttributeKeyAction)
 		if err != nil {
 			l.Error(":cold_sweat: Failed to get message action type with error: %s", err.Error())
 			continue
 		}
-		if messageType != "request" {
-			l.Debug(":ghost: Skipping non-request message type: %s", messageType)
-			continue
+
+		if messageType == otypes.EventTypeRequest {
```
sorawit: This should be matched again (MsgRequestData{}).Type

chain/cmd/bandoracled2/handler.go
```diff
@@ -25,17 +26,25 @@ func handleTransaction(c *Context, l *Logger, tx tmtypes.TxResult) {
...
+			go handleRequestLog(c, l, log)
+		} else if messageType == (ibc.MsgPacket{}).Type() {
+			// Try to get request id from packet. If not then return error.
+			_, err := GetEventValue(log, otypes.EventTypeRequest, otypes.AttributeKeyID)
+			if err != nil {
+				l.Debug(":ghost: Skipping non-request packet type: %s", messageType)
```
sorawit: I don't think we need %s, messageType here. because it will always be ics/opaque

chain/x/oracle/types/events.go
```diff
@@ -30,4 +30,5 @@ const (
 	AttributeKeyResolveStatus  = "resolve_status"
 	AttributeKeyResult         = "result"
 	AttributeKeyResultHash     = "result_hash"
+	AttributeKeyAction         = "action"
```
sorawit: Doesn't look right

sorawit (changes requested):

sorawit (approved):

taobun (approved):

PR #1689, opened 2020-05-21
oracled2: Log level

### LINKED ISSUE #1680: Allow user to configure log level in bandd
See how tendermint does this. User should be able to specify global flag `--log_level debug`. The default should be info.

perimeko:
Fixed #1680

### DIFF
--- a/chain/cmd/bandoracled2/run.go
+++ b/chain/cmd/bandoracled2/run.go
@@ -77,11 +77,9 @@ func runCmd(c *Context) *cobra.Command {
 			if err != nil {
 				return err
 			}
-			var logLevel string
+			logLevel := cfg.LogLevel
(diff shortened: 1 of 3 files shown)


chain/cmd/bandoracled2/run.go
```diff
@@ -76,8 +77,12 @@ func runCmd(c *Context) *cobra.Command {
 			if err != nil {
 				return err
 			}
-
-			l := NewLogger()
+			logLevel := cfg.LogLevel
```
sorawit: why do we need this temporary variable?

sorawit (changes requested):

sorawit (approved):

taobun (approved):

PR #1804, opened 2020-05-29
chain: Obi go lib test coverage 💯

### LINKED ISSUE #1718: Add unit test for OBI go lib

perimeko:
Fixed #1718


chain/pkg/obi/schema_test.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+type EmptySchema struct{}
+
+type Color struct {
+	code string `code` //missing obi
```
sorawit: Why is this not `obi:"code"`? Also `Code` should be public.

chain/pkg/obi/schema_test.go
```diff
@@ -0,0 +1,83 @@
...
+}
```
sorawit, line 13: Can we also add test for `bytes`?

sorawit (changes requested):

chain/pkg/obi/schema_test.go
perimeko: Added test case for bytes

perimeko: This struct is expected to be mock missing OBI tag struct.

chain/pkg/obi/schema_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+type EmptySchema struct{}
+
+type NoOBITagStruct struct {
+	NoOBITag string `"noOBItag"` //missing obi
+}
+
+type Validation struct {
```
sorawit: let's call this NotSupportedStruct

sorawit (changes requested):

sorawit (approved):

taobun (approved):

PR #1879, opened 2020-06-05
chain: Keep request and report data on chain state

sorawit:


chain/x/oracle/endblock.go
```diff
@@ -13,6 +13,8 @@ func handleEndBlock(ctx sdk.Context, k Keeper) {
 	}
 	// Once all the requests are resolved, we can clear the list.
 	k.SetPendingResolveList(ctx, []types.RequestID{})
-	// Lastly, we clean up old data requests from the primary storage.
+	// Lastly, we clean up data requests that are supposed to be expired.
 	k.ProcessExpiredRequests(ctx)
+	// TODO: We can remove old requests from storage to optimize state space, using `k.DeleteRequest`
```
taobun, line 18: I think this comment can be in ProcessExpiredRequests

sorawit: I think it should be here. Most likely we will want to keep requests around even after expired, and delete it after long period of time (say a month or so)

taobun (approved):

Benzbeeb (approved):

PR #1910, opened 2020-06-08
chain: Implement request-search REST and CLI endpoint

sorawit:


chain/x/oracle/client/common/request_search.go
```diff
@@ -0,0 +1,62 @@
...
+	route string, cliCtx context.CLIContext, oid, calldata, askCount, minCount string,
+) ([]byte, int64, error) {
+
+	events := []string{
+		fmt.Sprintf("%s.%s='%s'", types.EventTypeRequest, types.AttributeKeyOracleScriptID, oid),
+		fmt.Sprintf("%s.%s='%s'", types.EventTypeRequest, types.AttributeKeyCalldata, calldata),
+		fmt.Sprintf("%s.%s='%s'", types.EventTypeRequest, types.AttributeKeyAskCount, askCount),
+		fmt.Sprintf("%s.%s='%s'", types.EventTypeRequest, types.AttributeKeyMinCount, minCount),
+	}
+	searchResult, err := authclient.QueryTxsByEvents(cliCtx, events, 1, 30, "desc")
```
smiled0g, line 30: Why not use page 0? And if the script has multiple dApp using it (for example, median price oracle), will the limit of 30 be enough -- given that keys can be 100s of symbols requested for DEXes?

smiled0g: Ref: https://github.com/cosmos/cosmos-sdk/blob/master/x/auth/client/query.go

smiled0g (changes requested):

chain/x/oracle/client/common/request_search.go
sorawit: Because page 0 will lead to error
Screen Shot 2020-06-09 at 11 09 20

30 should be enough. It's filtered by oracle script id, calldata, mincount, and askcount. Bad situations can only happen there are 30 exact same requests waiting to be resolved, which is extremely unlikely given current parameter configurations.

taobun (dismissed):

smiled0g (approved):

taobun (approved):

PR #1931, opened 2020-06-11
Implement generate genesis with data source/oracle scripts flow 

### LINKED ISSUE #1884: Implement generate genesis with data source/oracle scripts flow

Benzbeeb:
Fixed #1884


owasm/chaindebugtests/moresilly/out.txt
```diff
@@ -0,0 +1 @@
+{symbol:string,multiplier:u64}/{px:u64}
```
taobun: Remove this file

chain/scripts/add_os_ds.sh
```diff
@@ -0,0 +1,188 @@
+bandd add-data-source \
```
taobun, line 2: I think an absolute path didn't make sense when need to compile on other device.

DIR=`dirname "$0"`

and use path as. $DIR/../datasources/ instead same as oracle script

taobun (changes requested):

PR #1939, opened 2020-06-11
Scan: use `Format` on user balance

### LINKED ISSUE #1488: Use Format.re instead of toString
1. Let's use `Format.iPretty` and `Format.fPretty` instead of `Js.Float.toString` or `string_of_float`.
2. Add `digits` argument defines how long digits must be.

evilpeach:
Fixes #1488


tansawit (approved):

tansawit (changes requested):
Please fix changelog conflict

prin-r (approved):

tansawit (approved):

PR #1963, opened 2020-06-13
Add bandchain.js

smiled0g:
Note: the old `bandchain.js` branch got messed up due to an incorrect rebase. This PR is for the new branch with the same content.


evilpeach (commented):
Should we add the test to GitHub action?

smiled0g:
> Should we add the test to GitHub action?

Hmm.. I'm not entirely sure here. I guess we should, but in that case, we should also use mock server requests instead of the real requests to GuanYu devnet we're doing right now.

What do you think? @sorawit

sorawit:
Yes. I think if we are to add it to github action, better be tests that don't rely on external API calls.

evilpeach (approved):

prin-r (approved):

helpers/bandchain.js/src/lib/BandChain.js
```diff
@@ -0,0 +1,195 @@
...
+    )
+
+    let signedTx = cosmos.sign(requestMsg, ecpairPriv, 'block')
+    convertSignedMsg(signedTx)
+
+    const broadcastResponse = await cosmos.broadcast(signedTx)
+    return this.getRequestID(broadcastResponse.txhash)
+  }
+
+  async getRequestID(txHash) {
```
prin-r: I saw retryTimeout = 200 at "getRequestProof". So should we add it here as well ?

taobun (changes requested):
Nope

helpers/README.md
```diff
@@ -0,0 +1,12 @@
+# BandChain Helpers
+
+A collection of tools that simplify interactions with BandChain. Note that this is not a part of BandChain standard and subject to change more often.
+
+## Supported Programming Languages
+
+| Language                        | Status         |
+| ------------------------------- | -------------- |
+| [`Python 3 🐍`](pyobi)          | 🤔 Not Started |
```
taobun: obi LUL

taobun (approved):

PR #1986, opened 2020-06-15
scan: add autofocus to submittx modal

### LINKED ISSUE #1980: Add auto focus on input (Submit Tx)
When the Submit Tx modal is popped up, the input should be focused.

tansawit:
Fixes: #1980


scan/src/components/modal/submitTx/SendMsg.re
```diff
@@ -70,6 +70,7 @@ let make = (~address, ~receiver, ~setMsgsOpt) => {
       msg="Recipient Address"
       code=true
       placeholder="Insert recipient address"
+      autoFocus=true
```
evilpeach: If the modal is opened with the recipient address, then focus on the amount's input.
If not, then focus on the recipient address input.

evilpeach (changes requested):

evilpeach (approved):

prin-r (approved):

PR #1988, opened 2020-06-16
Implement safe guard to check memory limit from wasm

### LINKED ISSUE #1965: Implement safe guard to check memory limit from wasm
- [ ] Implement function to verify memory have been allocated correctly.
- [ ] Add test that check we can reject wasm that allocate too much memory.
- [ ] But if wasm use dynamic memory make sure runtime can terminate after exceeding memory usage

Benzbeeb:
Fixed #1965


go-owasm/src/lib.rs
```diff
@@ -48,14 +49,38 @@ pub extern "C" fn do_wat2wasm(input: Span, output: &mut Span) -> Error {
     }
 }
 
+fn check_wasm_memories(module: &Module) -> Result<&Module, Error> {
+    let section = match module.memory_section() {
+        Some(section) => section,
+        None => return Err(Error::NoMemoryWasmError),
+    };
+
+    let memory = section.entries()[0];
```
sorawit, line 65: Is it always the case that `section.entries()` has size 1?

sorawit (changes requested):

sorawit (approved):
looks great

sorawit (approved):
looks great

taobun (approved):

PR #2001, opened 2020-06-18
Scan: binding count up, use it!

### LINKED ISSUE #1218: Binding and use React-Countup on account index page
- Use [React-countup](https://github.com/glennreyes/react-countup)
- [Binding Example](https://github.com/bandprotocol/bandx/blob/master/staking-reason/src/bindings/Countup.re)

Talk to @prin-r what position to apply this animation.

evilpeach:
Fixes #1218 

Using on:
- Validator Index Page (Delegate section)
- Withdraw Reward Modal
- Account Index Page


scan/src/components/NumberCountup.re
```diff
@@ -0,0 +1,26 @@
...
+  React.useEffect1(
+    _ => {
+      Countup.updateGet(countup, value);
+      None;
+    },
+    [|value|],
+  );
+
+  let newVal = Countup.countUpGet(countup);
+  <Text value={newVal |> Js.Float.toString} size weight spacing code=true nowrap=true />;
```
sorawit: Why don't we use fPretty here?

sorawit (changes requested):

scan/src/components/NumberCountup.re
evilpeach: I've changed to string type because the `CountUp.js` format the value for us.

sorawit (approved):

tansawit (approved):

PR #2005, opened 2020-06-18
scan: show commission max rate and change on validator index page

### LINKED ISSUE #1969: Add  min_commission_change, max_commission on Validator Index Page

tansawit:
Fixes: #1969


scan/src/subscriptions/ValidatorSub.re
```diff
@@ -70,6 +76,8 @@ let toExternal =
   details,
   tokens,
   commission: commissionRate *. 100.,
+  commissionMaxChange,
+  commissionMaxRate,
```
evilpeach: Should they multiple by 100 here?

evilpeach (changes requested):

scan/src/subscriptions/ValidatorSub.re
tansawit: I multiplied it by 100 when using it in ValidatorIndexPage before. But I've moved it to the sub to match the commission rate now.

evilpeach (changes requested):
Let's use `Format.fPercent(~digits=2)` on commission rate, commission max change, commission max rate and uptime.

evilpeach (approved):

prin-r (approved):

PR #2022, opened 2020-06-20
emitter/flusher: Initial implementation

sorawit:
Initial implementation of emitter/flusher.

Emitter is an add-on to BandApp to publish state changes to remote Kafka. Currently messages are flushed on every Commit. 

Flusher is a separate Python process that subscribes to messages the push changes to remote databases. It maintains the offset of Kafka messages to make sure no messages are missed.

This is POC implementation. More performance optimizations and features will be added in the future.


chain/emitter/app.go
```diff
@@ -0,0 +1,219 @@
...
+		if res.IsOK() {
+			extra = app.handleMsg(txHash, msg, logs[idx])
+		}
+		messages = append(messages, JsDict{
+			"msg":   msg,
+			"type":  msg.Type(),
+			"extra": extra,
+		})
+	}
+	txDict["messages"] = messages
```
Benzbeeb, line 189: d

flusher/flusher/sync.py
```diff
@@ -0,0 +1,57 @@
...
+                value = json.loads(msg.value)
+                if key == "COMMIT":
+                    if value["height"] % commit_interval == 0:
+                        conn.execute(tracking.update().values(kafka_offset=msg.offset))
+                        logger.info(
+                            "Committed at block {} and Kafka offset {}", value["height"], msg.offset
+                        )
+                        break
+                    continue
+                getattr(handler, "handle_" + key.lower())(json.loads(msg.value))
```
tansawit: Why not just use value? No need to do json.loads twice.

tansawit (changes requested):

chain/emitter/app.go
```diff
@@ -0,0 +1,219 @@
...
+	invCheckPeriod uint, skipUpgradeHeights map[int64]bool, home string,
+	baseAppOptions ...func(*bam.BaseApp),
+) *App {
+	return &App{
+		BandApp: bandapp.NewBandApp(
+			logger, db, traceStore, loadLatest, invCheckPeriod, skipUpgradeHeights,
+			home, baseAppOptions...,
+		),
+		writer: kafka.NewWriter(kafka.WriterConfig{
+			Brokers:      []string{"localhost:9092"}, // TOD: Remove hardcode
```
taobun: TODO

chain/emitter/handler.go
```diff
@@ -0,0 +1,29 @@
...
+	evMap := make(EvMap)
+	for _, event := range log.Events {
+		for _, kv := range event.Attributes {
+			key := event.Type + "." + kv.Key
+			evMap[key] = append(evMap[key], kv.Value)
+		}
+	}
+	switch msg := msg.(type) {
+	case oracle.MsgRequestData:
+		app.handleMsgRequestData(txHash, msg, evMap, extra)
```
taobun, line 21: Why does this function choose to send extra instead of return extra like handleMsg?

taobun (changes requested):

chain/emitter/handler.go
sorawit: So we don't need to create a new JsDict instance everywhere.

taobun (approved):

Benzbeeb (approved):

tansawit (approved):

PR #2027, opened 2020-06-21
obi/go: Make Encode and Decode able to work on multiple values

sorawit:


Benzbeeb (approved):

perimeko (approved):

chain/x/oracle/keeper/request.go
```diff
@@ -4,10 +4,12 @@ import (
 	"fmt"
 
 	"github.com/bandprotocol/bandchain/chain/pkg/bandrng"
```
taobun: Let's move to bottom

chain/x/oracle/types/packets_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"encoding/hex"
 	"testing"
 
+	"github.com/bandprotocol/bandchain/chain/pkg/obi"
```
taobun: Let's make it consistence

taobun (changes requested):

taobun (approved):

PR #2054, opened 2020-06-23
hotfix 🔥: patch scan for cosmos v38

tansawit:


evilpeach (changes requested):
It can't decode `account_number` on Wenchang version.

taobun (approved):

evilpeach (approved):

PR #2065, opened 2020-06-25
Remove obi.js build process

smiled0g:
To simplify the `obi.js` and make it work better with Node.js tools. We're removing the build process and use vanilla js instead. This should be fully compatible with all the tools that rely on `obi.js`.


tansawit (approved):

sorawit (approved):
Thanks!

PR #2066, opened 2020-06-25
Scan: comment out proposed block tab 

evilpeach:


evilpeach:
No need this anymore because we've created index on DB.

PR #2068, opened 2020-06-25
Remove band-consumer from repo

taobun:
We downgraded cosmos-sdk to version 0.38 and IBC module has been removed, so we don't need example chain to connect with Bandchain anymore.


sorawit (approved):
thanks!

evilpeach (approved):

PR #2074, opened 2020-06-25
chain: Use rolling block hash as seed for validator sampling

sorawit:


chain/pkg/bandrng/sampling.go
```diff
@@ -34,60 +28,45 @@ func ChooseOne(rng *Rng, weights []uint64) int {
...
-
-	for round := 0; round < k; round++ {
-		candidateSize := GetCandidateSize(round, k, totalCount)
-		luckyNumber := ChooseOne(rng, weightValues[:candidateSize])
-		luckies = append(luckies, weightIndexes[luckyNumber])
-
-		weightIndexes = append(weightIndexes[:luckyNumber], weightIndexes[luckyNumber+1:]...)
-		weightValues = append(weightValues[:luckyNumber], weightValues[luckyNumber+1:]...)
+// ChooseSomeMaxWeight performs ChooseSome "tries" times and returns the sampling with the
+// height weight sum among all tries.
```
taobun: highest?

chain/x/oracle/keeper/keeper.go
```diff
@@ -65,6 +65,21 @@ func (k Keeper) GetParams(ctx sdk.Context) (params types.Params) {
...
+func (k Keeper) SetRollingSeed(ctx sdk.Context, rollingSeed []byte) {
+	ctx.KVStore(k.storeKey).Set(types.RollingSeedStoreKey, rollingSeed)
+}
+
+// GetRollingSeed returns the current rolling seed value.
+func (k Keeper) GetRollingSeed(ctx sdk.Context) []byte {
+	bz := ctx.KVStore(k.storeKey).Get(types.RollingSeedStoreKey)
+	if bz == nil {
+		// If RollingSeedStoreKey is not yet set, we initialize it to a zero 32-byte slice.
+		return make([]byte, 32)
```
taobun: sha256.Size?

taobun (changes requested):

chain/pkg/bandrng/sampling.go
sorawit: Good catch!

chain/x/oracle/keeper/keeper.go
sorawit: I added a constant for this. It is 32, but it's not because it is sha256.size.

taobun (approved):

Benzbeeb (approved):

PR #2122, opened 2020-07-02
Scan: support ledger on window by using WebHID

evilpeach:
## Implementation details
- Use WebHID to support Ledger on Windows
- Implement disconnect function


scan/src/utils/Os.re
```diff
@@ -0,0 +1,18 @@
+let checkHID: unit => bool = [%bs.raw {|
+function() {
+  return !!navigator.hid;
+}
+|}];
+
+let isWindow: unit => bool = [%bs.raw
```
tansawit: change to `isWindows`?

tansawit (changes requested):

scan/src/utils/Os.re
evilpeach: Done!

sorawit (approved):
looks great

tansawit (approved):

PR #2139, opened 2020-07-03
chain: Remove db in favor of emitter

sorawit:
🗑


Benzbeeb:
> 🗑

We cannot erase the past, but we can accept it as history.

Benzbeeb (approved):

taobun (approved):

PR #2319, opened 2020-07-28
flusher: Add index on blocks table and swap order of primary key of validator_votes table

taobun:
## Implementation details
- Add index on some column that makes query slow.

## Please ensure the following
 requirements are met before submitting a pull request:

- [x] The pull request is targeted against the correct target branch
- [x] The pull request includes a description of the implementation/work done in detail.
- [x] You have added a relevant changelog entry to `CHANGELOG_UNRELEASED.md`
- [x] You have re-reviewed the files affected by the pull request (e.g. using the `Files changed` tab in the Github PR explorer)

## If this PR doesn't need to be merged into v1.0.6 please remove this option:

- [ ] The pull request has been cherry-picked to `rc/1.0.6`


flusher/flusher/db.py
```diff
@@ -248,13 +248,13 @@ def Column(*args, **kwargs):
 validator_votes = sa.Table(
     "validator_votes",
     metadata,
+    Column("block_height", sa.Integer, sa.ForeignKey("blocks.height"), primary_key=True),
```
sorawit, line 251: I think what we need to do here is to force order of the primary key columns.

sorawit (changes requested):

sorawit (approved):
thanks

PR #2602, opened 2020-09-02
Add docker multistage build

voron:
SSIA
Multistage build saves space, I see image size decrease from 1.47GB to 279MB
```
$ docker images
REPOSITORY               TAG                 IMAGE ID            CREATED             SIZE
bandd                    0                   e0ffd84ad3b7        26 minutes ago      279MB
blockchainetl/bandd      latest              58c698b214c2        About an hour ago   1.47GB
```


chain/Dockerfile
```diff
@@ -10,4 +10,16 @@ COPY chain/docker-config/run.sh .
...
 RUN make install && make faucet
 
+FROM debian:buster
+
+ENV WORKDIR=/chain
+WORKDIR ${WORKDIR}
+
+COPY --from=builder /go-owasm /go-owasm
+COPY --from=builder /chain/docker-config ${WORKDIR}/docker-config
+COPY --from=builder /chain/scripts ${WORKDIR}/scripts
```
taobun: I think we don't need this folder to the image.

taobun (changes requested):

chain/Dockerfile
voron: Just to be sure, are you talking about `/chain/scripts` directory ?

taobun: Yes, it's just a folder that contains convenience scripts for development.

voron: Got it, it's removed now

PR #2643, opened 2020-09-15
pyband: update reports and result type can be null

taobun:
## Implementation details
The report and result of the request can be null, so `get_request_by_id` can throw an exception if the field doesn't exist.

## Please ensure the following requirements are met before submitting a pull request:

- [x] The pull request is targeted against the correct target branch
- [x] The pull request includes a description of the implementation/work done in detail.
- [ ] You have added a relevant changelog entry to `CHANGELOG_UNRELEASED.md`
- [x] You have re-reviewed the files affected by the pull request (e.g. using the `Files changed` tab in the Github PR explorer)

### DIFF
--- a/CHANGELOG_UNRELEASED.md
+++ b/CHANGELOG_UNRELEASED.md
@@ -28,4 +28,6 @@
 
 ### Helpers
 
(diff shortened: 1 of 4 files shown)


sorawit (approved):
looks good

PR #2904, opened 2020-11-18
Add prometheus metrics for yoda

rav-11:
Adding prometheus metrics for yoda:

handling - number of reports currently being handled by the report executer
pending - number of reports waiting to be submitted to the chain
submitted - number of reports submitted (since the last restart of yoda)
error - number of errors logged by yoda


chain/yoda/execute.go
```diff
@@ -121,6 +121,8 @@ func SubmitReport(c *Context, l *Logger, keyIndex int64, reports []ReportMsgWith
 			switch txRes.Code {
 			case 0:
 				l.Info(":smiling_face_with_sunglasses: Successfully broadcast tx with hash: %s", txHash)
+				c.pendingGauge = c.pendingGauge - int64(len(reports))
+				c.submittedCount++
```
taobun: I think we want to track submitted reports, so we must add the length of reports?

chain/yoda/main.go
```diff
@@ -27,16 +27,18 @@ const (
...
+	NodeURI           string `mapstructure:"node"`                // Remote RPC URI of BandChain node to connect to
+	Validator         string `mapstructure:"validator"`           // The validator address that I'm responsible for
+	GasPrices         string `mapstructure:"gas-prices"`          // Gas prices of the transaction
+	LogLevel          string `mapstructure:"log-level"`           // Log level of the logger
+	Executor          string `mapstructure:"executor"`            // Executor name and URL (example: "Executor name:URL")
+	BroadcastTimeout  string `mapstructure:"broadcast-timeout"`   // The time that Yoda will wait for tx commit
+	RPCPollInterval   string `mapstructure:"rpc-poll-interval"`   // The duration of rpc poll interval
+	MaxTry            uint64 `mapstructure:"max-try"`             // The maximum number of tries to submit a report transaction
+	MaxReport         uint64 `mapstructure:"max-report"`          // The maximum number of reports in one transaction
+	Metrics           bool   `mapstructure:"metrics"`             // Switch for prometheus exporter
```
taobun: I think adding only 1 field and check if this variable has been set (not equal to an empty string) instead of a bool variable?

chain/yoda/context.go
```diff
@@ -47,6 +47,10 @@ type Context struct {
 
 	dataSourceCache *sync.Map
 	pendingRequests map[types.RequestID]bool
+
+	handlingGauge  int64
```
taobun, line 52: I'm not sure if these variables need to be threadsafe (using `sync/atomic`)? `handleRawRequest` for examples can be called concurrently by multiple goroutines.

chain/yoda/context.go
```diff
@@ -47,6 +47,10 @@ type Context struct {
...
+	pendingGauge   int64
+	submittedCount int64
 }
```
taobun, line 57: I think create increase function for each variable (increaseHandlingGauge, increasependingGauge, increasesubmittedCount) and update when flag is set(dont' do the atomic add if other node don't run the prometheus)

chain/yoda/run.go
```diff
@@ -46,6 +47,11 @@ func runImpl(c *Context, l *Logger) error {
 		return err
 	}
 
+	if cfg.MetricsListenAddr != "" {
```
taobun: set 1 more variable in context for check if need to track report

taobun (changes requested):

rav-11:
Have run this on our testnet server for a day and it's caught some errors and behaved as expected. Updated our mainnet server and will check behaviour is as expected.

rav-11:
After running for a week, it's worked as expected, and caught a couple of errors too. Ready for review again

chain/yoda/handler.go
```diff
@@ -218,27 +218,31 @@ func handleRawRequests(c *Context, l *Logger, id types.RequestID, reqs []rawRequ
 }
 
 func handleRawRequest(c *Context, l *Logger, req rawRequest, key keys.Info, id types.RequestID, processingResultCh chan processingResult) {
+	c.updateHandlingGauge(1)
```
taobun, line 221: Can we use defer c.updateHandlingGauge(-1) for this use case?

chain/yoda/execute.go
```diff
@@ -68,7 +68,8 @@ func SubmitReport(c *Context, l *Logger, keyIndex int64, reports []ReportMsgWith
 
 	for i, report := range reports {
 		if err := report.msg.ValidateBasic(); err != nil {
-			l.Error(":exploding_head: Failed to validate basic with error: %s", err.Error())
+			l.Error(":exploding_head: Failed to validate basic with error: %s", c, err.Error())
+			c.updatePendingGauge(int64(-len(reports)))
```
taobun: defer at top of a function?

chain/yoda/handler.go
rav-11: Updated

chain/yoda/execute.go
rav-11: Updated

rav-11:
@taobun Updated as requested, and run updated version on our testnet and mainnet servers for a few minutes with no issue

chain/yoda/handler.go
```diff
@@ -218,10 +218,12 @@ func handleRawRequests(c *Context, l *Logger, id types.RequestID, reqs []rawRequ
 }
 
 func handleRawRequest(c *Context, l *Logger, req rawRequest, key keys.Info, id types.RequestID, processingResultCh chan processingResult) {
+	defer c.updateHandlingGauge(-1)
```
taobun: I think +1 should come before defer

rav-11: Updated

taobun (approved):

